PR #2, opened 2019-01-28
Abstract exchanging keying material

backkem:
- Add an abstraction to exchange keying material between the DTLS
  and SRTP package.
- Construct & start the Session at the same time to avoid races.

This makes `RTCDtlsTransport.startSRTP()` look as follows:
```go
keys, err := srtp.ExtractSessionKeysFromDTLS(t.conn, t.isClient())
if err != nil {
	return fmt.Errorf("failed to extract sctp session keys: %v", err)
}

srtpConfig := &srtp.Config{
	Keys:    keys,
	Profile: srtp.ProtectionProfileAes128CmHmacSha1_80,
}

srtpSession, err := srtp.NewSessionSRTP(t.srtpEndpoint, srtpConfig)
if err != nil {
	return fmt.Errorf("failed to start srtp: %v", err)
}

srtcpSession, err := srtp.NewSessionSRTCP(t.srtcpEndpoint, srtpConfig)
if err != nil {
	return fmt.Errorf("failed to start srtcp: %v", err)
}

t.srtpSession = srtpSession
t.srtcpSession = srtcpSession
```

### DIFF
--- a/keying.go
+++ b/keying.go
@@ -1,0 +1,45 @@
+package srtp
+
+const labelExtractorDtlsSrtp = "EXTRACTOR-dtls_srtp"
+
+// KeyingMaterialExporter allows package SRTP to extract keying material
+type KeyingMaterialExporter interface {
+	ExportKeyingMaterial(label string, context []byte, length int) ([]byte, error)
+}
+
+// ExtractSessionKeysFromDTLS allows setting the Config SessionKeys by
+// extracting them from DTLS. This behavior is defined in RFC5764:
+// https://tools.ietf.org/html/rfc5764
+func (c *Config) ExtractSessionKeysFromDTLS(exporter KeyingMaterialExporter, isClient bool) error {
+	keyingMaterial, err := exporter.ExportKeyingMaterial(labelExtractorDtlsSrtp, nil, (keyLen*2)+(saltLen*2))
+	if err != nil {
+		return err
+	}
+
+	offset := 0
+	clientWriteKey := append([]byte{}, keyingMaterial[offset:offset+keyLen]...)
+	offset += keyLen
+
+	serverWriteKey := append([]byte{}, keyingMaterial[offset:offset+keyLen]...)
+	offset += keyLen
+
+	clientWriteKey = append(clientWriteKey, keyingMaterial[offset:offset+saltLen]...)
+	offset += saltLen
+
+	serverWriteKey = append(serverWriteKey, keyingMaterial[offset:offset+saltLen]...)
+
+	if isClient {
+		c.Keys.LocalMasterKey = clientWriteKey[0:keyLen]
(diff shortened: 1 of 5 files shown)


keying.go
```diff
@@ -0,0 +1,47 @@
...
+// KeyingMaterialExporter allows package SRTP to extract keying material
+type KeyingMaterialExporter interface {
+	// TODO: To match crypto/tls label should be a string.
+	ExportKeyingMaterial(label []byte, context []byte, length int) ([]byte, error)
+}
+
+// ExtractSessionKeysFromDTLS extracts the session keys from DTLS
+// This behavior is defined in RFC5764
+// https://tools.ietf.org/html/rfc5764
+func ExtractSessionKeysFromDTLS(exporter KeyingMaterialExporter, isClient bool) (SessionKeys, error) {
```
Sean-Der: We need `ProtectionProfile` in here somewhere, the length/offsets are determine by that (and we will need to support that in the future) don't want to break the API when that happens.

keying.go
```diff
@@ -0,0 +1,47 @@
...
+	offset += keyLen
+
+	clientWriteKey = append(clientWriteKey, keyingMaterial[offset:offset+saltLen]...)
+	offset += saltLen
+
+	serverWriteKey = append(serverWriteKey, keyingMaterial[offset:offset+saltLen]...)
+
+	if isClient {
+		return SessionKeys{
+			LocalMasterKey:   clientWriteKey[0:16],
```
Sean-Der: Can you use the `keyLen` constants here? When we support more `ProtectionProfile`s will be helpful

keying.go
```diff
@@ -0,0 +1,47 @@
+package srtp
+
+// KeyingMaterialExporter allows package SRTP to extract keying material
+type KeyingMaterialExporter interface {
+	// TODO: To match crypto/tls label should be a string.
+	ExportKeyingMaterial(label []byte, context []byte, length int) ([]byte, error)
```
Sean-Der: Bleh, mind fixing this TODO? Even if you don't update `DTLS` we can have a little wrapper that does `[]byte -> string` before calling `ExportKeyingMaterial` I am afraid little tech debt like this will make people ignore `pions/dtls`

session_srtcp.go
```diff
@@ -18,18 +18,29 @@ type SessionSRTCP struct {
 	writeStream *WriteStreamSRTCP
 }
 
-// CreateSessionSRTCP creates a new SessionSRTCP
-func CreateSessionSRTCP() *SessionSRTCP {
-	s := &SessionSRTCP{}
+// NewSessionSRTCP creates a SRTCP session using conn as the underlying transport.
+func NewSessionSRTCP(conn net.Conn, config *Config) (*SessionSRTCP, error) {
+	s := &SessionSRTCP{
```
Sean-Der: Nice! This should fix the reported race (hopefully)

PR #5, opened 2019-02-14
 Speed up EncryptRTP

kixelated:
**new:**
```
BenchmarkEncryptRTP-8          	  300000	      5181 ns/op	    2512 B/op	      11 allocs/op
BenchmarkEncryptRTPInPlace-8   	  300000	      4480 ns/op	    1104 B/op	      10 allocs/op
```

**old:**
```
BenchmarkEncryptRTP-8          	  200000	      6378 ns/op	    4304 B/op	      12 allocs/op
BenchmarkEncryptRTPInPlace-8   	  300000	      4448 ns/op	    1104 B/op	      10 allocs/op
```

The old code would:

1. `allocateIfMismatch` performs `bytes.Equal` to check if the dst/src
are the same. It's not needed, it could technically have false-positives,
and it could have been done faster with pointers anyway.
    ex: `&dst[0] == &src[0]`

2. `allocateIfMismatch` would always copy the payload, even if we're
just about to overwrite it with encrypted data.

3. `allocateIfMismatch` will make a byte slice of the size/cap of the
payload slice, but won't include extra room for the auth tag (10 bytes)
we're expecting to append. This means that we will always cause an extra
realloc/copy, as evidenced by the benchmark memstats.

The unit test to check if the encryption was done in place was broken.
The ciphertext was being written to the dst, but then append was used
to write the auth tag. This causes a reallocation unless the dst is
allocated with 10 extra bytes of overhead. So encrypting in place would
often not work.

I switched the API to explicitly append instead of write to [0]. The
stdlib likes to switch kind of arbitrarily between both approaches. If
the output size is not trivial, then it will always append. But if the
output size is fixed, then it can write in place. When the output size
can be calculated (like `len(plaintext)+10`), most APIs still use append.
For example, cipher.Block writes to indexes directly while cipher.AEAD
will append because it also attaches an auth tag.

Writing to indexes directly should always error if `len(dst)` is not large
enough, as a way to guarantee against reallocations. The current approach
uses append under the hood so it's better to be explicit that there may be
a reallocation.

You can still append to the same slice by using `dst[:0]`. It will set
the len to 0 but the cap will stay the same.

### DIFF
--- a/srtp.go
+++ b/srtp.go
@@ -44,36 +44,62 @@ func (c *Context) DecryptRTP(dst, encrypted []byte, header *rtp.Header) ([]byte,
 	return c.decryptRTP(dst, encrypted, header)
 }
 
-func (c *Context) encryptRTP(dst, decrypted []byte, header *rtp.Header) ([]byte, error) {
-	dst = allocateIfMismatch(dst, decrypted)
-
-	s := c.getSSRCState(header.SSRC)
-
-	c.updateRolloverCount(header.SequenceNumber, s)
-
-	stream := cipher.NewCTR(c.srtpBlock, c.generateCounter(header.SequenceNumber, s.rolloverCounter, s.ssrc, c.srtpSessionSalt))
-	stream.XORKeyStream(dst[header.PayloadOffset:], dst[header.PayloadOffset:])
-
-	dst = append(dst, make([]byte, 4)...)
-	binary.BigEndian.PutUint32(dst[len(dst)-4:], s.rolloverCounter)
-
-	authTag, err := c.generateAuthTag(dst, c.srtpSessionAuthTag)
-	if err != nil {
-		return nil, err
-	}
-
-	return append(dst[:len(dst)-4], authTag...), nil
-}
-
-// EncryptRTP Encrypts a RTP packet
-func (c *Context) EncryptRTP(dst, decrypted []byte, header *rtp.Header) ([]byte, error) {
+// EncryptRTP encrypts a plaintext RTP packet, writing to the dst buffer provided.
+// If the dst buffer does not have the capacity to hold `len(plaintext) + 10` bytes, a new one will be allocated.
+// If a rtp.Header is provided, it will be Unmarshaled using the plaintext.
+func (c *Context) EncryptRTP(dst []byte, plaintext []byte, header *rtp.Header) ([]byte, error) {
+	// TODO(@lcurley) Potentially accept a *rtp.Packet to avoid this Unmarshal
 	if header == nil {
 		header = &rtp.Header{}
 	}
 
-	if err := header.Unmarshal(decrypted); err != nil {
+	err := header.Unmarshal(plaintext)
+	if err != nil {
 		return nil, err
 	}
 
-	return c.encryptRTP(dst, decrypted, header)
+	// Write to dst starting at this offset.
+	offset := 0
+
+	// Grow the given buffer to fit the output.
+	// authTag = 10 bytes
+	dst = growBufferSize(dst, len(plaintext)+10)
+
+	s := c.getSSRCState(header.SSRC)
+	c.updateRolloverCount(header.SequenceNumber, s)
+
+	// Copy the header unencrypted.
+	headerSize := header.PayloadOffset
+
+	// Skip the copy if the two slices share memory addresses.
+	if &dst[0] != &plaintext[0] {
+		copy(dst[offset:], plaintext[:headerSize])
+	}
+
+	offset += headerSize
+
+	// Encrypt the payload
+	counter := c.generateCounter(header.SequenceNumber, s.rolloverCounter, s.ssrc, c.srtpSessionSalt)
+	stream := cipher.NewCTR(c.srtpBlock, counter)
+	stream.XORKeyStream(dst[offset:], plaintext[headerSize:])
+	offset += len(plaintext) - headerSize
+
+	// Write the rollover counter just for computing the auth tag.
+	// We will overwrite it immediately afterwards.
+	binary.BigEndian.PutUint32(dst[offset:], s.rolloverCounter)
+	offset += 4
+
+	// Generate the auth tag.
+	authTag, err := c.generateAuthTag(dst[:offset], c.srtpSessionAuthTag)
+	if err != nil {
+		return nil, err
+	}
+
+	// Pop off the rollover counter.
+	offset -= 4
+
+	// Write the auth tag to the dest.
+	copy(dst[offset:], authTag)
+
+	return dst, nil
 }
--- a/srtp_test.go
+++ b/srtp_test.go
@@ -122,23 +122,22 @@ func TestRolloverCount(t *testing.T) {
 	}
 }
 
-func buildTestContext(t *testing.T) *Context {
+func buildTestContext() (*Context, error) {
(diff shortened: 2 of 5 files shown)


srtp.go
```diff
@@ -44,36 +44,63 @@ func (c *Context) DecryptRTP(dst, encrypted []byte, header *rtp.Header) ([]byte,
 	return c.decryptRTP(dst, encrypted, header)
 }
 
-func (c *Context) encryptRTP(dst, decrypted []byte, header *rtp.Header) ([]byte, error) {
-	dst = allocateIfMismatch(dst, decrypted)
+// EncryptRTP Encrypts a RTP packet
+// This method modifies rtp.Header based on the contents of plaintext.
+// TODO This function should take (*rtp.Packet) instead of ([]byte, rtp.Header)
```
Sean-Der: We designed the API so that people who are just trying to decrypt packets can use an API that makes sense for them (resembling libsrtp). I don't know if this is a change we will make.

srtp.go
```diff
@@ -44,36 +44,63 @@ func (c *Context) DecryptRTP(dst, encrypted []byte, header *rtp.Header) ([]byte,
...
+// TODO Don't modify *rtp.Header; ideally just remove it.
```
Sean-Der: This is on purpose. Users can pass in a `*rtp.Header` if they have already unmarshalled and we just modify it for them.

If they pass nil we allocate a new one

stream_srtp.go
```diff
@@ -116,13 +116,13 @@ type WriteStreamSRTP struct {
 }
 
 // WriteRTP encrypts a RTP header and its payload to the nextConn
-func (w *WriteStreamSRTP) WriteRTP(header *rtp.Header, payload []byte) (int, error) {
-	headerRaw, err := header.Marshal()
+func (w *WriteStreamSRTP) WriteRTP(packet *rtp.Packet) (int, error) {
```
Sean-Der: This API is on purpose, all of the `pion/` transports do the same thing.

srtp_test.go
```diff
@@ -244,13 +257,13 @@ func TestRTPLifecyleInPlace(t *testing.T) {
 		}
 
 		// Copy packet, asserts that everything was done in place
-		encryptInput := make([]byte, len(decryptedRaw))
+		encryptInput := make([]byte, len(decryptedRaw), len(decryptedRaw)+10)
 		copy(encryptInput, decryptedRaw)
 
-		actualEncrypted, err := encryptContext.EncryptRTP(encryptInput, encryptInput, encryptHeader)
+		actualEncrypted, err := encryptContext.EncryptRTP(encryptInput[:0], encryptInput, encryptHeader)
```
Sean-Der: Why `encryptInput[:0]`? Can we do just `encryptInput`, people will look at the tests to understand how this works and might be confused by that.

This code tries to mock the [AEAD interface](https://golang.org/pkg/crypto/cipher/#AEAD) I probably messed up in a lot of places though

Sean-Der:
Fantastic work @kixelated!

Write in place is a poor choice, but I was testing for a realloc (and maybe my assumptions were wrong) but I would hope that people could pass the buffer and we could allocate the tag for them.

Or they could allocation ahead of time (we should add a case to test for this also!) nice little improvement there.

I see three unique inputs though (and would be nice to test for them all)

* New allocation
* realloc for auth tag
* completely in place

srtp_test.go
kixelated: AEAD will append to the buffer, not write directly to it. In fact:

```
// To reuse plaintext's storage for the encrypted output, use plaintext[:0]
// as dst. Otherwise, the remaining capacity of dst must not overlap plaintext.
```

Basically you have to decide if the API should start writing bytes at index `0` or `len(dst)`. `cipher.Stream` does the former, while `cipher.AEAD` and `hash.Hash` does the latter. I can't think of an API that both starts at index `0` and uses `append` to expand the buffer like this implementation. 

I'm actually kind of fine not changing the API in the sake of backwards compatibility. I mostly don't like passing in the `*rtp.Header` and I'm kind of indifferent to the dst different as it only matters when `len(dst) != 0`

srtp.go
kixelated: Yeah, I see how you're trying to make generic by taking a byte slice. I do think it was a good goal but it has a performance impact.

Take this example:

```
raw, _ := packet.Marshal()
raw, _ = EncryptRTP(raw, raw, nil)
```

It will actually cause a reallocation, because `raw` won't have the capacity for the extra 10 bytes. I don't expect the caller of the function to realize that the function is not actually encrypting in place.

To avoid this, you have the use the shiny new `MarshalTo`. It's something that I would only expect `EncryptRTP` to do because it knows the auth size.

```
// Allocate 10 extra bytes for the auth tag.
// Otherwise it will cause a reallocation.
raw := make([]byte, 0, packet.MarshalSize() + 10)
raw, _ = packet.MarshalTo(raw)
raw, _ = EncryptRTP(raw, raw, nil)
```

We currently have to `Unmarshal` for header information, immediately after we just called `Marshal`. It's not a huge performance hit but it's something. Hell, I actually expected the point of passing in a `rtp.Header` was to avoid doing this `Unmarshal`, but nope.

I do agree that a 3rd party should be able to encrypt a `[]byte` slice. I also think that a 3rd party should be able to encrypt a `*rtp.Packet`. So one of the two has to `Marshal` or `Unmarshal` before calling `EncryptRTP`.

current:
```
// WriteStreamSRTP.WriteRTP and ones that deal with rtp.Packet
raw, _ = packet.Marshal()
raw, _ = EncryptRTP(raw, raw, nil) // REALLOCATION

// other libraries that deal with raw []byte slices
raw, _ = EncryptRTP(raw, raw, nil) // REALLOCATION
```

proposed:
```
// WriteStreamSRTP.WriteRTP and ones that deal with rtp.Packet
raw, _ = EncryptRTP(raw, packet)

// other libraries that deal with raw []byte slices
p, _ := rtp.UnmarshalPacket(raw)
raw, _ = EncryptRTP(raw, raw) // REALLOCATION
```

Truth be told, it's a pretty minor gripe that requires a lot of refactoring. Backwards compatibility makes more sense.

srtp.go
kixelated: I don't understand. The raw packet hasn't changed, so neither will `*rtp.Header` right? Maybe I'm missing the part of the code that modifies the packet.

It only seems useful if you have a raw packet and don't want to `Unmarshal` it until after you've sent it again... and you want to save nanoseconds.

stream_srtp.go
kixelated: I'm a little too eager to make breaking changes for no reason in this case. 

However, it means I'm going to have to use `rtp.MarshalSize()` and `rtp.MarshalTo()` to optimize this code, because it's doing an extra allocation for no reason.

kixelated:
@Sean-Der Thanks for the review and sorry for being so eager to break the API. I'm going to revert the API changes and just improve the performance.

Sean-Der:
Hey @kixelated 

I think it is awesome you want to break the API actually :) your fresh insight is SUPER valuable, and it would be a huge mistake if I tempered it by accident.

Everything you think is wrong please put on the wiki page, I want to do them all. I just have to balance stability and making everyone happy is my only concern :( and I could be doing a bad job of that so please always push back if I am wrong.
-----

Tell me when you are ready for another review! If there are no API changes that would be fantastic, we can bump in pions/webrtc and keep getting these wins :) really excited about the promise of faster ciphers

kixelated:
Should be good for a review now that the linter passes.

kixelated:
oh wait except for `stream_srtp.go`

PR #6, opened 2019-02-16
Optimize session auth tag generation

kixelated:
I profiled my application and crypto/hmac.New was taking 5.71% of the
total CPU time. Since the session key does not change, you can call
crypto/hmac.Reset() instead of making a new object. This reuses previous
work done for the key.

```
name                 old time/op  new time/op  delta
EncryptRTP-8         2.17µs ± 4%  1.90µs ± 5%  -12.79%  (p=0.008 n=5+5)
EncryptRTPInPlace-8  2.14µs ± 4%  1.84µs ± 4%  -13.79%  (p=0.008 n=5+5)
```

### DIFF
--- a/context.go
+++ b/context.go
@@ -5,8 +5,8 @@ import (
 	"crypto/cipher"
 	"crypto/hmac"
 	"crypto/sha1" // #nosec
-	"crypto/subtle"
 	"encoding/binary"
+	"hash"
 
 	"github.com/pkg/errors"
 )
@@ -56,11 +56,13 @@ type Context struct {
 	ssrcStates         map[uint32]*ssrcState
 	srtpSessionKey     []byte
 	srtpSessionSalt    []byte
+	srtpSessionAuth    hash.Hash
 	srtpSessionAuthTag []byte
 	srtpBlock          cipher.Block
 
 	srtcpSessionKey     []byte
 	srtcpSessionSalt    []byte
+	srtcpSessionAuth    hash.Hash
 	srtcpSessionAuthTag []byte
 	srtcpIndex          uint32
 	srtcpBlock          cipher.Block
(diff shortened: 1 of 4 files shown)


context.go
```diff
@@ -199,7 +207,32 @@ func (c *Context) generateCounter(sequenceNumber uint16, rolloverCounter uint32,
...
+	// The pre-defined authentication transform for SRTP is HMAC-SHA1
+	// [RFC2104].  With HMAC-SHA1, the SRTP_PREFIX_LENGTH (Figure 3) SHALL
+	// be 0.  For SRTP (respectively SRTCP), the HMAC SHALL be applied to
+	// the session authentication key and M as specified above, i.e.,
+	// HMAC(k_a, M).  The HMAC output SHALL then be truncated to the n_tag
+	// left-most bits.
+	// - Authenticated portion of the packet is everything BEFORE MKI
+	// - k_a is the session message authentication key
+	// - n_tag is the bit-length of the output authentication tag
+	// - ROC is already added by caller (to allow RTP + RTCP support)
```
Sean-Der: Since this is all split up now can you update it so that the ROC is passed into `generateSrtpAuthTag`, just so we don't have to do as much slicing (adding and then removing it)

Also can you update the comments, `ROC is already added by the...` not a really helpful point anymore.

kixelated: Yeah, this cleans up the code a little bit; I like it.

kixelated:
btw, `DecryptRTCP` doesn't check the auth tag.

Sean-Der:
@kixelated 

Blah thanks for catching that, would be awesome if you could fix + add tests for that also! Thanks so much

context.go
```diff
@@ -214,27 +221,42 @@ func (c *Context) generateAuthTag(buf, sessionAuthTag []byte) ([]byte, error) {
...
+	c.srtpSessionAuth.Reset()
 
-	if _, err := mac.Write(buf); err != nil {
+	if _, err := c.srtpSessionAuth.Write(buf); err != nil {
 		return nil, err
 	}
 
-	return mac.Sum(nil)[0:10], nil
+	// For SRTP only, we need to hash the rollover counter.
+	rocRaw := []byte{byte(roc >> 24), byte(roc >> 16), byte(roc >> 8), byte(roc)}
```
Sean-Der: How much of a performance loss/gain do we have from this bit shifting? I would prefer things be explicit, as much as the performance matters we will eventually get crushed when people don't want to touch/contribute I think.

kixelated: It just saves needing to clear the memory to zero, so basically no performance difference. I was trying to be clever and save a line of code.

PR #8, opened 2019-02-17
Optimize WriteRTP

kixelated:
The previous code would Marshal the header allocating (header) bytes.
Then call append with the payload allocating payload bytes and copying
(header) bytes. Then later EncryptRTP will append the auth tag,
potentially allocating (header + payload + 10) bytes and copying
(header + payload) bytes.

This code avoids the extra allocation/copy by allocating the correct
size upfront.

```
name                 old time/op    new time/op    delta
WriteRTP-8             9.73µs ± 2%    9.33µs ± 2%   -4.11%

name                 old alloc/op   new alloc/op   delta
WriteRTP-8             3.17kB ± 0%    1.10kB ± 0%  -65.15%

name                 old allocs/op  new allocs/op  delta
WriteRTP-8               13.0 ± 0%      10.0 ± 0%  -23.08%
```

The benchmark prints four `io: read/write on closed pipe` log lines but I wasn't able to figure it how to fix them. It's pretty annoying but I'll fix it later.

Depends on https://github.com/pions/rtp/pull/3


stream_srtp.go
```diff
@@ -113,17 +113,32 @@ func (r *ReadStreamSRTP) GetSSRC() uint32 {
 // WriteStreamSRTP is stream for a single Session that is used to encrypt RTP
 type WriteStreamSRTP struct {
 	session *SessionSRTP
+	buffer  []byte
```
Sean-Der: Is this thread safe? `WriteStreamSRTP` all of the `Stream` stuff should be.

kixelated: It's not thread-safe. I wasn't sure if it was a requirement, but sure I can change it.

Sean-Der: yea if you don't mind!

I think we need to make everything thread safe by default, just to avoid the sharp edges. I made that mistake at first and it bit a lot of people :(

kixelated:
This change destructively overwrites the buffer, which is not supposed to be done by `io.Writer`s. Alongside the need to make the buffer reuse thread-safe, there's really no performance improvement any longer. Just one less allocation the size of the header so really not worth the effort to do anything but CLOSE.

PR #11, opened 2019-02-22
Fix out-of-bounds slice access in DecryptRTCP

### LINKED ISSUE #10: Out-of-bounds access in DecryptRTCP
On line 14, `out` is truncated to length `tailOffset`:
https://github.com/pions/srtp/blob/bc58b55b1a0d101267701ae3ea00412cb77bc918/srtcp.go#L14

However, on line 21, access of `out[tailOffset:]` is attempted, which results in an out-of-bounds panic:
https://github.com/pions/srtp/blob/bc58b55b1a0d101267701ae3ea00412cb77bc918/srtcp.go#L21

Believe line 21 should be:

    srtcpIndexBuffer := encrypted[tailOffset : tailOffset+srtcpIndexSize]

thinkski:
Slice was truncated to exclude the authentication portion, but was
then accessed beyond the truncated portion to read the index. Test
did not catch it because slice backing memory happened to be valid.

Replaced unsetting of 'encrypted' bit in memory with a bitmask on
index, as that is all it was needed for. With only the above fix,
tests would now fail. Both changes are necessary for tests to pass.

Fixes #10

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Check out the **[contributing wiki](https://github.com/pions/webrtc/wiki/Contrib
 * [Tobias Fridén](https://github.com/tobiasfriden) *SRTP authentication verification*
 * [chenkaiC4](https://github.com/chenkaiC4) - *Fix GolangCI Linter*
 * [Luke Curley](https://github.com/kixelated) - *Performance*
+* [Chris Hiszpanski](https://github.com/thinkski) - *Fix out-of-bounds access*
 
 ### License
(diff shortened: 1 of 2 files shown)


Sean-Der:
Oof nice find, thank you @thinkski 

Mind adding yourself to the README, but everything else LGTM!

thinkski:
> Oof nice find, thank you @thinkski
> 
> Mind adding yourself to the README, but everything else LGTM!

Of course, it would be my pleasure :)

PR #13, opened 2019-03-01
Use batched network writes if supported

kixelated:
Will use batching if net.Conn also implements:
    `WriteBatch(packets [][]byte) (n int, err error)`
    `ReadBatch(packets [][]byte) (n int, err error)`

In this case, `n` is the number of packets read/written, which can be
less than `len(packets)`. This matches how sendmmsg behaves.

In order to support writing multiple packets at once, we either need a
new function or modify the existing API. I'm really not a fan of
`WriteRTP(payload []byte, header *rtp.Packet) (n int, err error)`.

I've replaced it with `WriteRTP(packets ...*rtp.Packet) error` instead
of making a new function. This is a breaking API change but matches the
new webrtc API.

Alongside this change, I have a new version of `encryptRTP` that takes
an `rtp.Packet` instead of `[]byte plaintext`. When used in the
`WriteRTP` pipeline, it avoids an extra header Unmarshal and an extra
payload Marshal (aka copy).

note: This compares new `encryptRTP` and old `EncryptRTP`, which is only
used for `WriteRTP` and not `Write`.

```
EncryptRTP-8         1.93µs ± 1%  1.85µs ± 5%  -3.93%
EncryptRTPInPlace-8  1.90µs ± 3%  1.80µs ± 7%  -5.31%
WriteRTP-8           5.29µs ± 2%  5.28µs ± 1%    ~
```


go.sum
```diff
@@ -2,6 +2,8 @@ github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/pions/rtcp v1.0.0 h1:kYGe6RegZ63yVDkqXaru1+kHZAqHEufP3zfRAGKPycI=
 github.com/pions/rtcp v1.0.0/go.mod h1:Q5twXlqiz775Yn37X0cl4lAsfSk8EiHgeNkte59jBY4=
+github.com/pions/rtp v1.0.0 h1:H/TUg7bhgBT/mQsUx0adW3cmgwqPmygoYbbRTc3Y7Ek=
```
Sean-Der: `go mod tidy` 

Do we need this old version, or just not upgraded locally?

go.mod
```diff
@@ -5,4 +5,5 @@ require (
 	github.com/pions/rtp v1.0.1
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.3.0
+	golang.org/x/net v0.0.0-20190227160552-c95aed5357e7
```
Sean-Der: Do we need this new dependency? I couldn't find an import

go.sum
kixelated: Probably not. We at least need `MarshalTo`.

connection.go
```diff
@@ -0,0 +1,51 @@
+package srtp
```
Sean-Der: nice! is this something that can live in `github.com/pions/transport` ? Do you see other subsystems using it (like SCTP) ?

srtp.go
```diff
@@ -54,45 +54,60 @@ func (c *Context) DecryptRTP(dst, encrypted []byte, header *rtp.Header) ([]byte,
...
 
-	s := c.getSSRCState(header.SSRC)
-	c.updateRolloverCount(header.SequenceNumber, s)
+	s := c.getSSRCState(packet.Header.SSRC)
+	c.updateRolloverCount(packet.Header.SequenceNumber, s)
 
 	// Copy the header unencrypted.
-	headerSize := header.PayloadOffset
+	// The dst slicing is weird because MarshalTo uses append while others write directly.
+	rawHeader, err := packet.Header.MarshalTo(dst[offset:offset])
```
kixelated: I'm actually with changing `MarshalTo` now that I've tried to use it a bit. The append functionality is okay but it doesn't match `io.Reader` like most things pions uses.

The one caveat is that it should error if `len(data) < packet.MarshalSize()`. The old code would append and accidentally create copies. It also means the API would be:

`packet.MarshalTo(data []byte) (n int, err error)` aka `io.Reader`

srtp.go
```diff
@@ -54,45 +54,60 @@ func (c *Context) DecryptRTP(dst, encrypted []byte, header *rtp.Header) ([]byte,
...
-// If a rtp.Header is provided, it will be Unmarshaled using the plaintext.
-func (c *Context) EncryptRTP(dst []byte, plaintext []byte, header *rtp.Header) ([]byte, error) {
-	// TODO(@lcurley) Potentially accept a *rtp.Packet to avoid this Unmarshal
-	if header == nil {
-		header = &rtp.Header{}
-	}
+// EncryptRTP marshals and encrypts an RTP packet, writing to the dst buffer provided.
+// If the dst buffer does not have the capacity to hold `len(plaintext) + 10` bytes, a new one will be allocated and returned.
+// If the header is provided, it will be unmarshaled using the plaintext.
+func (c *Context) EncryptRTP(dst []byte, plaintext []byte, header *rtp.Header) (ciphertext []byte, err error) {
```
kixelated: I would like to replace with API with `EncryptRTP(dst []byte, packet *rtp.Packet) (ciphertext []byte, err error)` as seen below. It's not critical so I didn't make the change.

kixelated:
Depends on https://github.com/pions/transport/pull/5

kixelated:
Turns out that it's a performance decrease.

kixelated:
I'm going to pull the crypto changes out of this code review for that performance boost though.

PR #14, opened 2019-03-05
Optimize WriteRTP

kixelated:
The old EncryptRTP API causes an extra header Unmarshal and payload
Marshal. By passing the `*rtp.Packet` through the stack, we're able to
avoid both of these.

There was also an extra reallocation caused by using append.

```
name                 old time/op  new time/op  delta
EncryptRTP-8         1.94µs ± 3%  1.94µs ± 3%    ~
EncryptRTPInPlace-8  1.89µs ± 3%  1.89µs ± 4%    ~
DecryptRTP-8         1.80µs ± 3%  1.83µs ± 7%    ~
Write-8              1.95µs ± 8%  1.93µs ± 1%   ~
WriteRTP-8           2.11µs ± 6%  2.01µs ± 8%  -4.71%
```

### DIFF
--- a/srtp.go
+++ b/srtp.go
@@ -54,11 +54,10 @@ func (c *Context) DecryptRTP(dst, encrypted []byte, header *rtp.Header) ([]byte,
 	return c.decryptRTP(dst, encrypted, header)
 }
 
-// EncryptRTP encrypts a plaintext RTP packet, writing to the dst buffer provided.
-// If the dst buffer does not have the capacity to hold `len(plaintext) + 10` bytes, a new one will be allocated.
+// EncryptRTP marshals and encrypts an RTP packet, writing to the dst buffer provided.
+// If the dst buffer does not have the capacity to hold `len(plaintext) + 10` bytes, a new one will be allocated and returned.
 // If a rtp.Header is provided, it will be Unmarshaled using the plaintext.
 func (c *Context) EncryptRTP(dst []byte, plaintext []byte, header *rtp.Header) ([]byte, error) {
-	// TODO(@lcurley) Potentially accept a *rtp.Packet to avoid this Unmarshal
 	if header == nil {
 		header = &rtp.Header{}
 	}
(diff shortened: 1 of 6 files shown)


stream_srtp.go
```diff
@@ -115,15 +115,9 @@ type WriteStreamSRTP struct {
 	session *SessionSRTP
 }
 
-// WriteRTP encrypts a RTP header and its payload to the nextConn
-func (w *WriteStreamSRTP) WriteRTP(header *rtp.Header, payload []byte) (int, error) {
```
kixelated:
```
         .          .    119:func (w *WriteStreamSRTP) WriteRTP(header *rtp.Header, payload []byte) (int, error) {
         .      200ms    120:	headerRaw, err := header.Marshal()
         .          .    121:	if err != nil {
         .          .    122:		return 0, err
         .          .    123:	}
         .          .    124:
         .          .    125:	// TODO(@lcurley) This will cause one, potentially two, extra allocations.
         .      230ms    126:	raw := append(headerRaw, payload...)
      40ms      9.40s    127:	return w.session.write(raw)
         .          .    128:}
```

srtp.go
```diff
@@ -54,45 +54,38 @@ func (c *Context) DecryptRTP(dst, encrypted []byte, header *rtp.Header) ([]byte,
 	return c.decryptRTP(dst, encrypted, header)
 }
 
-// EncryptRTP encrypts a plaintext RTP packet, writing to the dst buffer provided.
-// If the dst buffer does not have the capacity to hold `len(plaintext) + 10` bytes, a new one will be allocated.
-// If a rtp.Header is provided, it will be Unmarshaled using the plaintext.
-func (c *Context) EncryptRTP(dst []byte, plaintext []byte, header *rtp.Header) ([]byte, error) {
-	// TODO(@lcurley) Potentially accept a *rtp.Packet to avoid this Unmarshal
```
kixelated:
```
         .          .     62:	if header == nil {
         .       10ms     63:		header = &rtp.Header{}
         .          .     64:	}
         .          .     65:
         .       40ms     66:	err := header.Unmarshal(plaintext)
         .          .     67:	if err != nil {
         .          .     68:		return nil, err
         .          .     69:	}
```

This weird header unmarshalling is less of a performance bottleneck but it's something.

Sean-Der:
Hey @kixelated merge w/e you are ready this LGTM!

PR #16, opened 2019-03-05
Optimize getOrCreateReadStream

kixelated:
The allocation of a readStream every time was causing a ton of
allocations for no reason. Passing in a function avoids the
allocation.

### DIFF
--- a/session.go
+++ b/session.go
@@ -45,7 +45,7 @@ type SessionKeys struct {
 	RemoteMasterSalt []byte
 }
 
-func (s *session) getOrCreateReadStream(ssrc uint32, child streamSession, proto readStream) (readStream, bool) {
+func (s *session) getOrCreateReadStream(ssrc uint32, child streamSession, proto func() readStream) (readStream, bool) {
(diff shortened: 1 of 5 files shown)


kixelated:
I don't think the anonymous function improves readability. I'm not sure about the performance impact either but I imagine it wouldn't have one.

The main problem with the old code is that the compiler had to allocate the reader on the heap, meanwhile this function pointer (and possibly your anonymous function) can be allocated on the stack. Go performs escape analysis to determine if a given variable can escape the current scope, which was true only when creating a new readStream.

Sean-Der:
Readability wise

* newReadStreamSRTCP is a new function I have to keep track of when reading/understanding the code. So I have to jump to it when reading and realize it does nothing.

* newReadStreamSRTCP is something that other people may use, and can create code that is interwoven more then it needs to be and people might modify it and create side effects

kixelated:
I mean it's a constructor. I don't think people should be confused that `newReadStreamSRTCP` creates a new `*ReadStreamSRTCP`.

PR #25, opened 2019-04-07
Upgrade to pion/rtcp v1.1.5

maxhawkins:


maxhawkins:
https://github.com/pion/srtp/blob/master/session_srtcp.go#L117

This is going to need to be updated to support compound packets. Reader doesn't support CompoundPacket right now.

Issue #26, opened 2019-04-07
Upgrade to new rtcp package


maxhawkins:

Sean-Der:
This is done. SRTP now uses `v1.2.0` (the latest release of pion/rtcp)

PR #27, opened 2019-04-11
Use new optional RTCP validation

wdouglass:
against compound packets

Relates to #26

### DIFF
--- a/session_srtcp.go
+++ b/session_srtcp.go
@@ -106,6 +106,25 @@ func (s *SessionSRTCP) write(buf []byte) (int, error) {
 	return s.session.nextConn.Write(encrypted)
 }
 
+//create a list of Destination SSRCs
+//that's a superset of all Destinations in the slice.
+func destinationSSRC(pkts []rtcp.Packet) []uint32 {
+	ssrcSet := make(map[uint32]struct{})
+	for _, p := range pkts {
+		for _, ssrc := range p.DestinationSSRC() {
+			ssrcSet[ssrc] = struct{}{}
+		}
+	}
+
+	out := make([]uint32, len(ssrcSet))
+
+	for ssrc := range ssrcSet {
+		out = append(out, ssrc)
+	}
+
+	return out
+}
+
 func (s *SessionSRTCP) decrypt(buf []byte) error {
 	decrypted, err := s.remoteContext.DecryptRTCP(buf, buf, nil)
 	if err != nil {
@@ -117,7 +136,7 @@ func (s *SessionSRTCP) decrypt(buf []byte) error {
 		return err
 	}
 
-	for _, ssrc := range pkt.DestinationSSRC() {
+	for _, ssrc := range destinationSSRC(pkt) {
 		r, isNew := s.session.getOrCreateReadStream(ssrc, s, newReadStreamSRTCP)
 		if r == nil {
 			return nil // Session has been closed
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Chris Hiszpanski](https://github.com/thinkski) - *Fix out-of-bounds access*
 * [Yutaka Takeda](https://github.com/enobufs) - *Fix log messages*
 * [Max Hawkins](https://github.com/maxhawkins)
+* [Woodrow Douglass](https://github.com/wdouglass)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.12
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/pion/logging v0.2.1
-	github.com/pion/rtcp v1.1.5
+	github.com/pion/rtcp v1.2.0
 	github.com/pion/rtp v1.1.1
 	github.com/pion/transport v0.6.0
 	github.com/pkg/errors v0.8.1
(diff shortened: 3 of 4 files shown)


session_srtcp.go
```diff
@@ -112,28 +112,50 @@ func (s *SessionSRTCP) decrypt(buf []byte) error {
...
+			} else if isNew {
+				s.session.newStream <- r // Notify AcceptStream
+			}
+
+			readStream, ok := r.(*ReadStreamSRTCP)
+			if !ok {
+				return false, fmt.Errorf("failed to get/create ReadStreamSRTP")
+			}
+
+			_, err = readStream.write(decrypted)
```
maxhawkins: Confused why this is `decrypted`.

If a compound packet with `n` subpackets arrives won't this just deliver the compound packet `n` times?

session_srtcp.go
```diff
@@ -112,28 +112,50 @@ func (s *SessionSRTCP) decrypt(buf []byte) error {
...
-
-		_, err = readStream.write(decrypted)
+	cpkt := rtcp.CompoundPacket(pkt)
+	if cpkt.Validate() == nil {
+		_, err = forward(&cpkt)
 		if err != nil {
 			return err
 		}
+	} else {
+		for _, p := range pkt {
```
maxhawkins: Shouldn't an rtcp-rsize packet with multiple sub-packets should be delivered as one `[]Packet` rather than split up? Sending two packets in one datagram might have some semantic meaning in the application.

session_srtcp.go
wdouglass: you're right, and your other comment below is related. good catch! let me take another pass.

session_srtcp.go
wdouglass: hmmm... may have to take another look at the spec. do all of the packets in an rtcp-rsize packet share a set of DestinationSSRC?

maxhawkins: Good question. I believe the whole compound packet goes to the same SSRC but I'm not sure on the rsize ones.

wdouglass: i've just force-pushed a revision that sends the whole mess to a superset of `Packet.DestinationSSRC()` for each packet in an rtcp-rsize packet (that isn't also a valid `CompoundPacket`)

wdouglass: Also, test coverage for srtp is pretty bad, but i don't really have time to look at it today. I'm going to file an issue.

session_srtcp.go
```diff
@@ -112,25 +112,47 @@ func (s *SessionSRTCP) decrypt(buf []byte) error {
 		return err
 	}
 
-	pkt, err := rtcp.Unmarshal(decrypted)
-	if err != nil {
-		return err
-	}
-
-	for _, ssrc := range pkt.DestinationSSRC() {
+	forward := func(ssrc uint32) (closed bool, err error) {
```
maxhawkins: Don't think pulling this into a function adds clarity. It's confusing what `return true, nil` signifies, for instance.

It's only called once so it would probably be better inline like before.

session_srtcp.go
```diff
@@ -112,25 +112,47 @@ func (s *SessionSRTCP) decrypt(buf []byte) error {
...
+	}
+
+	pkt, err := rtcp.Unmarshal(decrypted)
+	if err != nil {
+		return err
+	}
+
+	cpkt := rtcp.CompoundPacket(pkt)
+	ssrcSet := make(map[uint32]bool)
+	if cpkt.Validate() == nil {
```
maxhawkins: Why validate here at all? Both ends of this if statement will have the same effect.

Let's let applications do compound packet validation instead.

session_srtcp.go
```diff
@@ -112,25 +112,47 @@ func (s *SessionSRTCP) decrypt(buf []byte) error {
...
+		for _, ssrc := range cpkt.DestinationSSRC() {
+			ssrcSet[ssrc] = true
+		}
+	} else {
+		for _, p := range pkt {
+			for _, ssrc := range p.DestinationSSRC() {
```
maxhawkins: I wasn't sure what sssrcSet was for at first. Maybe this could be pulled out into a function?

```go
for _, ssrc := range destinationSSRCs(packets) {
  ...
}
```

wdouglass: The first branch assumes that the DestinationSSRC for a whole CompoundPacket set applies to the whole set, the second branch creates a superset of all of the DestinationSSRC for each packet in the set

the behavior is slightly different (not sure if that's desireable)

session_srtcp.go
wdouglass: The first branch assumes that the DestinationSSRC for a whole CompoundPacket set applies to the whole set, the second branch creates a superset of all of the DestinationSSRC for each packet in the set

the behavior is slightly different (not sure if that's desireable)

session_srtcp.go
wdouglass: fair, i'm reverting that.

session_srtcp.go
maxhawkins: I think the second branch is the proper behavior.

Say you receive this compound packet indicating that streams with ssrc1, ssrc2, and ssrc3 are all ending:
&CompoundPacket{
  SR(ssrc1),
  SDES(ssrc1 = 'cname1'),
  BYE(ssrc1, ssrc2, ssrc3)
}

You would want to send that to all three srtcp streams.

session_srtcp.go
wdouglass: what about adding this function to the rtcp repository:

```
func DestinationSSRC(pkts []Packet) []uint32 {
	cpkt := CompoundPacket(pkts)
	if cpkt.Validate() == nil {
		//this is a CompoundPacket; the rules about where to send
		//it are pretty well-documented.
		return cpkt.DestinationSSRC()
	}

	ssrcSet := make(map[uint32]struct{})
	//this isn't a proper CompoundPacket; let's create a list of Destination SSRCs
	//that's a superset of all Destinations in the slice.
	for _, p := range pkts {
		for _, ssrc := range p.DestinationSSRC() {
			ssrcSet[ssrc] = struct{}{}
		}
	}

	out := make([]uint32, len(ssrcSet))

	for ssrc := range ssrcSet {
		out = append(out, ssrc)
	}

	return out
}
```

maxhawkins: I'd rather not grow the public API right now. If it's useful we can add it later.

session_srtcp.go
wdouglass: oh! ok, i'll remove the branch.

PR #30, opened 2019-05-04
Add OnRTCPReceived callback

mjmac:
With this commit plus a small modification to webrtc.RTPSender,
a user could easily set a callback to be invoked when RTCP packets are received.

If ya'll like this approach I'll add some tests to make it mergeable.

Proposed modification to webrtc.RTPSender:
```go
func (r *RTPSender) OnRTCPReceived(f func([]rtcp.Packet)) {
       go func() {
               select {
               case <-r.sendCalled:
                       r.rtcpReadStream.OnRTCPReceived(f)
               case <-r.stopCalled:
                       return
               }
       }()
 }
```

Example usage by user code:
```go
sender, err := peer.connection.AddTrack(track)
if err != nil {
        return emptySdp, errors.Wrapf(err, "Unable to add %s track to PeerConnection %s", sourceID, peerID)
}
sender.OnRTCPReceived(func(msg []rtcp.Packet) {
        logs.Debugf("PeerConnection %s stream %s received RTCP: %s", peerID, sourceID, msg)
        // TODO: Implement sender pacing based on stream feedback (REMB, NACKs, etc)
})
```

### DIFF
--- a/stream_srtcp.go
+++ b/stream_srtcp.go
@@ -13,10 +13,11 @@ const srtcpBufferSize = 100 * 1000
 
 // ReadStreamSRTCP handles decryption for a single RTCP SSRC
 type ReadStreamSRTCP struct {
-	mu sync.Mutex
+	mu sync.RWMutex
 
-	isInited bool
-	isClosed chan bool
+	isInited              bool
+	isClosed              chan bool
+	onRTCPReceivedHandler func([]rtcp.Packet)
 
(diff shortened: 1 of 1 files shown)


mjmac:
Closing after some discussion in the #pion channel. This approach is too specific to my use case. For consistency, we'd also need to add an OnRTP callback, and then we'd have two different ways to receive these messages.

PR #33, opened 2019-08-24
Remove pkg/errors dependency

Sean-Der:
Everything we used it for is provided by stdlib


daenney:
You can get to it for earlier Go versions through https://godoc.org/golang.org/x/xerrors

enobufs:
@backkem Looks like the change has dropped the use of Wrap() and switched to the simple fmt.Errorf() and t.Errorf().

backkem:
You're right, seems I was still sleeping earlier 
...

Issue #28, opened 2019-04-11
the 'session' objects are completely not covered by tests.


wdouglass:
This issue is a reminder for me to revisit these objects (but If anyone wants to write some tests before i get to it, feel free!)

PR #34, opened 2019-09-01
Cover SessionSRTP with basic tests

Sean-Der:
Add basic read/write/close testing

Relates to #28


session_srtp.go
```diff
@@ -53,12 +55,6 @@ func NewSessionSRTP(conn net.Conn, config *Config) (*SessionSRTP, error) {
 	return s, nil
 }
 
-// Start initializes any crypto context and allows reading/writing to begin
```
Sean-Der: This is actually started by initializing the session, we shouldn't expose this API anymore since it could cause double starts.

Sean-Der:
Ok I feel pretty good about this coverage! When this lands I will take another look at easy wins, but hard to tell with all the churn atm.

Issue #22, opened 2019-04-02
"full buffer" messages on console


mjmac:
### Your environment.
  - Version: v1.1.2/b2f6a7fc
  - Other Information:

I added some extra logging to help figure out why this message was appearing:
```diff
--- a/vendor/github.com/pions/transport/packetio/buffer.go
+++ b/vendor/github.com/pions/transport/packetio/buffer.go
@@ -1,9 +1,10 @@
 package packetio

 import (
-       "errors"
        "io"
        "sync"
+
+       "github.com/pkg/errors"
 )

 // ErrFull is returned when the buffer has hit the configured limits.
@@ -57,7 +58,7 @@ func (b *Buffer) Write(packet []byte) (n int, err error) {
        // Check if there is available capacity
        if b.limitSize != 0 && b.size+len(packet) > b.limitSize {
                b.mutex.Unlock()
-               return 0, ErrFull
+               return 0, errors.Wrap(ErrFull, "Buffer capacity reached in Write()")
        }
```

```diff
--- a/vendor/github.com/pions/srtp/session.go
+++ b/vendor/github.com/pions/srtp/session.go
@@ -5,6 +5,8 @@ import (
        "io"
        "net"
        "sync"
+
+       "github.com/pkg/errors"
 )

 type streamSession interface {
@@ -92,6 +94,10 @@ func (s *session) close() error {
        return nil
 }

+type stackTracer interface {
+       StackTrace() errors.StackTrace
+}
+
 func (s *session) start(localMasterKey, localMasterSalt, remoteMasterKey, remoteMasterSalt []byte, profile ProtectionProfile, child streamSession) error {
        var err error
        s.localContext, err = CreateContext(localMasterKey, localMasterSalt, profile)
@@ -127,6 +133,11 @@ func (s *session) start(localMasterKey, localMasterSalt, remoteMasterKey, remote

                        if err = child.decrypt(b[:i]); err != nil {
                                fmt.Println(err)
+                               if err, ok := err.(stackTracer); ok {
+                                       for _, f := range err.StackTrace() {
+                                               fmt.Printf("%+s:%d\n", f, f)
+                                       }
+                               }
                        }
                }
        }()
```

This gave me the following stack trace on the console (one of zillions, once they started appearing):
```
Buffer capacity reached in Write(): full buffer
savant-go/vendor/github.com/pions/transport/packetio.(*Buffer).Write
[...]

PR #53, opened 2020-01-30
Spamming info message into log

NovelCorpse:
#### Description
Hello, everyone! I've got same #22 problem with spamming info messages regarding full buffer into logs. 

I'm not sure, but may be we can fix it like for [stream_srtp](https://github.com/pion/srtp/blob/master/stream_srtp.go#L59) ?
Also I've removed unusable [condition](https://github.com/pion/srtp/blob/master/stream_srtcp.go#L56), because packetio.Buffer's Read() never returns ErrFull.

I would appreciate for your reply.
Best regard, Roman.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Woodrow Douglass](https://github.com/wdouglass)
 * [Hugo Arregui](https://github.com/hugoArregui)
 * [Atsushi Watanabe](https://github.com/at-wat)
+* [Novel Corpse](https://github.com/NovelCorpse)
 
 ### License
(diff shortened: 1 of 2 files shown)


Sean-Der:
@NovelCorpse I am fine with this change! Since we are doing it in the RTP path.

I bet it was just a mistake doing it on the Read (instead of Write) when CI passes I will merge :) would you mind adding yourself to the README?

thanks for using Pion!

NovelCorpse:
@Sean-Der I've updated README. CI passed, but codecov/patch still failing. Is it a problem? 😞

Sean-Der:
@NovelCorpse this is perfect! I will merge now and bump `pion/webrtc` to use your new version.

Thank you so much for your contribution :)

PR #58, opened 2020-03-09
Add replay protection

at-wat:
Add functional options to set replay protection window to the Context.

Disable funlen linter. Splitting TestSessionSRTPReplayProtection
badly breaks the single test logic to pieces.

### Reference issue
~Requires https://github.com/pion/transport/pull/61~ merged


context.go
```diff
@@ -58,27 +60,36 @@ type Context struct {
...
 	srtcpSessionAuth    hash.Hash
 	srtcpSessionAuthTag []byte
 	srtcpIndex          uint32
 	srtcpBlock          cipher.Block
+	srtcpReplayDetector replaydetector.ReplayDetector
 }
 
 // CreateContext creates a new SRTP Context
-func CreateContext(masterKey, masterSalt []byte, profile ProtectionProfile) (c *Context, err error) {
+func CreateContext(masterKey, masterSalt []byte, profile ProtectionProfile, opts ...ContextOption) (c *Context, err error) {
```
Sean-Der: I think this might be too generic/not self documenting. 

What do you think of adding this as either a method or a mandatory argument? I don't know if a developer consuming this package would understand what `ContextOption` means.

srtcp.go
```diff
@@ -20,6 +22,16 @@ func (c *Context) decryptRTCP(dst, encrypted []byte) ([]byte, error) {
...
 	}
 
+	srtcpIndexBuffer := encrypted[tailOffset : tailOffset+srtcpIndexSize]
+
+	index := binary.BigEndian.Uint32(srtcpIndexBuffer) &^ (1 << 31)
+	ssrc := binary.BigEndian.Uint32(encrypted[4:])
+
+	markAsValid, ok := c.srtcpReplayDetector.Check(uint64(index))
+	if !ok {
+		return nil, errDuplicated
```
Sean-Der: Do you know what the UX for `pion/webrtc` will be like right now? Does `mux` just print this error, or does it `Close`/tear down everything?

at-wat: This error is returned to the user code only if `srtp.Context` is directly used. `pion/webrtc` uses srtp via `srtp.SessionSRT(C)P`. `errDuplicated` is consumed at https://github.com/pion/srtp/blob/a61c2acb40ffcd747f4e72c06dce8fa658082d9b/session.go#L132-L134

session_srtp_test.go
```diff
@@ -138,3 +142,120 @@ func TestSessionSRTPOpenReadStream(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestSessionSRTPReplayProtection(t *testing.T) {
```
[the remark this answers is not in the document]
at-wat: I think we should confirm that replay attack packet on any timing is rejected. (at least in the protection window, outer the window, and counter wrapping boundary.)
Making it 10 good packets and 1 replay only confirm protection at single timing.

context.go
at-wat: `ContextOption` is following functional options pattern, so it basically generic (we can put any option changing `Context` private members) and not only for replay protection.

IMO, it's better to avoid breaking API at now.
There is another API breaking backlog. It would be better to break API at one time.

In my experience, enabling SRTP replay protection at default may cause breakage on receiving stream from independently running command like gst-launch, since such command randomly select initial sequence number. Reinitialized stream is discarded by replay protection.

at-wat:
@Sean-Der I added response to your comments. Please take another look when you have time.

Sean-Der:
`Merging #58 into master will increase coverage by 7.99%.` holy shit amazing.

at-wat:
I'll add a tag v1.3.0.

PR #62, opened 2020-03-25
Fix replay detector scope

### LINKED ISSUE #60: ReplayDetector is shared among multiple SSRCs
`srtp.Context` is shared among multiple SSRCs, but roll over counter and replay detector must not be shared.
Shared ROC causes decryption failure when the difference of actual ROCs is more than 1.
(Packet rate of each stream can be different.)
And, replay detector causes false positive detection.

ref: https://github.com/pion/webrtc/pull/1095#issuecomment-603169345

at-wat:
Hold replay detector per SSRC.

### Reference issue
Fixes #60 
Related https://github.com/pion/webrtc/pull/1095

### DIFF
--- a/context.go
+++ b/context.go
@@ -45,6 +45,7 @@ type ssrcState struct {
 	rolloverCounter      uint32
 	rolloverHasProcessed bool
 	lastSequenceNumber   uint16
+	replayDetector       replaydetector.ReplayDetector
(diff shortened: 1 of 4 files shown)


at-wat:
All tests of pion/webrtc passed with this fix.

PR #75, opened 2020-05-30
Reduce lock contention in (*SessionSRTP).writeRTP.

jech:
This avoids performing I/O under a lock.  On a 4-core CPU, this has been measured as reducing roughly by half the total amount of lock contention in the Unnamed SFU, not counting contention due to the operations queue.

The remaining lock contention is almost entirely at line 129 just above, but that's for later.


Sean-Der:
Fantastic :)

I merged with 65c27fe7f77f16c6c0125d4b07e6477a9a13b2e2, I made the same change in the SRTCP code.

PR #76, opened 2020-06-01
Optimise away calls to cipher.NewCTR

jech:
This avoids the allocation of a 512-byte buffer on each packet, which leads to a 30% to 50% speedup in the SRTP benchmaks.

Edit (28/04/2022): see https://github.com/pion/srtp/pull/76#issuecomment-1112323998 for updated benchmarks.

```
name                 old time/op    new time/op     delta
EncryptRTP-4           4.15µs ± 0%     2.85µs ± 1%  -31.38%  (p=0.000 n=8+7)
EncryptRTPInPlace-4    4.05µs ± 2%     2.75µs ± 0%  -32.19%  (p=0.000 n=7+8)
DecryptRTP-4           3.81µs ± 0%     2.02µs ± 1%  -47.01%  (p=0.000 n=8+8)
Write-4                4.23µs ± 1%     2.93µs ± 0%  -30.79%  (p=0.000 n=8+8)
WriteRTP-4             4.15µs ± 0%     2.86µs ± 1%  -31.10%  (p=0.000 n=7+8)

name                 old speed      new speed       delta
EncryptRTP-4         27.0MB/s ± 0%   39.3MB/s ± 1%  +45.71%  (p=0.000 n=8+7)
EncryptRTPInPlace-4  27.7MB/s ± 2%   40.8MB/s ± 0%  +47.46%  (p=0.000 n=7+8)
DecryptRTP-4         7.34MB/s ± 0%  13.86MB/s ± 1%  +88.73%  (p=0.000 n=8+8)
Write-4              26.5MB/s ± 1%   38.2MB/s ± 0%  +44.47%  (p=0.000 n=8+8)
WriteRTP-4           27.0MB/s ± 0%   39.2MB/s ± 1%  +45.16%  (p=0.000 n=7+8)

name                 old alloc/op   new alloc/op    delta
EncryptRTP-4             788B ± 0%       212B ± 0%  -73.10%  (p=0.000 n=8+8)
EncryptRTPInPlace-4      660B ± 0%        84B ± 0%  -87.27%  (p=0.000 n=8+8)
DecryptRTP-4             692B ± 0%       116B ± 0%  -83.24%  (p=0.000 n=8+8)
Write-4                  788B ± 0%       212B ± 0%  -73.10%  (p=0.000 n=8+8)
WriteRTP-4               788B ± 0%       212B ± 0%  -73.10%  (p=0.000 n=8+8)

name                 old allocs/op  new allocs/op   delta
EncryptRTP-4             7.00 ± 0%       6.00 ± 0%  -14.29%  (p=0.000 n=8+8)
EncryptRTPInPlace-4      6.00 ± 0%       5.00 ± 0%  -16.67%  (p=0.000 n=8+8)
DecryptRTP-4             7.00 ± 0%       6.00 ± 0%  -14.29%  (p=0.000 n=8+8)
Write-4                  7.00 ± 0%       6.00 ± 0%  -14.29%  (p=0.000 n=8+8)
WriteRTP-4               7.00 ± 0%       6.00 ± 0%  -14.29%  (p=0.000 n=8+8)

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -1,0 +1,58 @@
+package srtp
+
+import (
+	"crypto/cipher"
+)
+
+// xorBytes computes the exclusive-or of src1 and src2 and stores it in dst.
+// It returns the number of bytes written.
+func xorBytes(dst, src1, src2 []byte) int {
+	n := len(src1)
+	if len(src2) < n {
+		n = len(src2)
+	}
+	if len(dst) < n {
+		n = len(dst)
+	}
+
+	for i := 0; i < n; i++ {
+		dst[i] = src1[i] ^ src2[i]
+	}
+
+	return n
+}
+
+// incrementCTR increments a big-endian integer of arbitrary size.
+func incrementCTR(ctr []byte) {
+	for i := len(ctr) - 1; i >= 0; i-- {
+		ctr[i]++
+		if ctr[i] != 0 {
+			break
+		}
+	}
+}
+
+// xorBytesCTR performs CTR encryption and decryption.
+// It is equivalent to cipher.NewCTR followed by XORKeyStream.
+func xorBytesCTR(block cipher.Block, iv []byte, dst, src []byte) error {
+	if len(iv) != block.BlockSize() {
+		return errBadIVLength
+	}
+
+	ctr := make([]byte, len(iv))
+	copy(ctr, iv)
+	bs := block.BlockSize()
+	stream := make([]byte, bs)
+
+	i := 0
+	for i < len(src) {
+		block.Encrypt(stream, ctr)
+		incrementCTR(ctr)
+		n := xorBytes(dst[i:], src[i:], stream)
+		if n == 0 {
+			break
+		}
+		i += n
+	}
+	return nil
+}
--- a/crypto_test.go
+++ b/crypto_test.go
@@ -1,0 +1,159 @@
+package srtp
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func xorBytesCTRReference(block cipher.Block, iv []byte, dst, src []byte) {
+	stream := cipher.NewCTR(block, iv)
+	stream.XORKeyStream(dst, src)
+}
+
+func TestXorBytesCTR(t *testing.T) {
+	for keysize := 16; keysize < 64; keysize *= 2 {
+		key := make([]byte, keysize)
+		_, err := rand.Read(key) //nolint: gosec
+		require.NoError(t, err)
+
+		block, err := aes.NewCipher(key)
+		require.NoError(t, err)
+
+		iv := make([]byte, block.BlockSize())
+		for i := 0; i < 1500; i++ {
+			src := make([]byte, i)
+			dst := make([]byte, i)
+			reference := make([]byte, i)
+			_, err = rand.Read(iv) //nolint: gosec
+			require.NoError(t, err)
+
+			_, err = rand.Read(src) //nolint: gosec
+			require.NoError(t, err)
+
+			assert.NoError(t, xorBytesCTR(block, iv, dst, src))
+			xorBytesCTRReference(block, iv, reference, src)
+			require.Equal(t, dst, reference)
+
+			// test overlap
+			assert.NoError(t, xorBytesCTR(block, iv, dst, dst))
+			xorBytesCTRReference(block, iv, reference, reference)
+			require.Equal(t, dst, reference)
+		}
+	}
+}
+
+func TestXorBytesCTRInvalidIvLength(t *testing.T) {
+	key := make([]byte, 16)
+	block, err := aes.NewCipher(key)
+	require.NoError(t, err)
+
+	src := make([]byte, 1024)
+	dst := make([]byte, 1024)
+
+	test := func(iv []byte) {
+		assert.Error(t, errBadIVLength, xorBytesCTR(block, iv, dst, src))
+	}
+
+	test(make([]byte, block.BlockSize()-1))
+	test(make([]byte, block.BlockSize()+1))
+}
+
+func TestXorBytesBufferSize(t *testing.T) {
+	a := []byte{3}
+	b := []byte{5, 6}
+	dst := make([]byte, 3)
+
+	xorBytes(dst, a, b)
+	require.Equal(t, dst, []byte{6, 0, 0})
+
+	xorBytes(dst, b, a)
+	require.Equal(t, dst, []byte{6, 0, 0})
+
+	a = []byte{1, 1, 1, 1}
+	b = []byte{2, 2, 2, 2}
+	dst = make([]byte, 3)
+
+	xorBytes(dst, a, b)
+	require.Equal(t, dst, []byte{3, 3, 3})
+}
+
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: crypto.go [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+		ctr[i]++
+		if ctr[i] != 0 {
+			break
+		}
+	}
+}
+
+// xorBytesCTR performs CTR encryption and decryption.
+// It is equivalent to cipher.NewCTR followed by XORKeyStream.
+func xorBytesCTR(block cipher.Block, iv []byte, dst, src []byte) {
```
at-wat, line 37: Please add following error handlings to avoid future security problems. Having tests for error cases is better for the security.
- initial vector length, input buffer sizes, and block size
- validity of overwrapped buffer address (ones in the official crypto package: https://github.com/golang/go/blob/50bd1c4d4eb4fac8ddeb5f063c099daccfb71b26/src/crypto/cipher/ctr.go#L79-L81)

crypto.go
```diff
@@ -0,0 +1,50 @@
+package srtp
+
+import (
+	"crypto/cipher"
+)
+
+// xorBytes computes the exclusive-or of src1 and src2 and stores it in dst.
+// It returns the number of bytes written.
+func xorBytes(dst, src1, src2 []byte) int {
```
at-wat, line 9: Could you add test cases to ensure behavior against different combination of buffer length?

Sean-Der:
I vote for merging it, but I think we should keep it out of the default code path. I worry that a mistake in this code could really damage Pion's reputation. With DTLS we didn't have an option, but for this if we mess it up it could damage user trust for ever. People love to harp on `do not roll your own crypto` and even if the code isn't in the real crypto stuff they would use this as the example :/

What do you think of making this controllable via a new `ContextOption`? We can then put this in the `SettingEngine` in pion/webrtc.

I think we should open a bug upstream and see what they say. Hopefully we can land something like this upstream, and then put the new API behind a version check!

at-wat:
> `do not roll your own crypto`

I really thought it.
At least, such code must be tested for various possible (especially erroneous) input patterns.
Making this controllable via a new `ContextOption` would be good for now, I think.

jech:
> I vote for merging it, but I think we should keep it out of the default code path

I understand your hesitation, but I'm not sure I agree.

This is not "rolling your own crypto".  I'm just reimplementing a well-known, standard, boring piece of crypto machinery that has seen extensive peer review and that is very well understood.  This is completely different from designing new crypto algorithms.  (If we cannot even reimplement well-known algorithms, then why does Go have its own crypto stack rather than using OpenSSL?)

Second, any issue will be readily apparent: there is a test that it produces the same values for all reasonable packet sizes as the reference implementation.

Third, any security issues in this code will be readily apparent even without the above test — it's not anything subtle like checking MACs or protecting against replay, it's just stream encryption.  If something goes wrong notwithstanding the test above, then the symptoms will be incorrectly encrypted or decrypted packets which will be dropped.

> What do you think of making this controllable via a new `ContextOption`?

Oh, that's overkill.  If for some reason you want to make it user controllable, just define a global variable that can be initialised in `main`.

> I think we should open a bug upstream and see what they say.

 You do it or I do?

crypto.go
jech: * iv length: done;
* input buffer size: this is CTR, it works with arbitrary sizes;
* block size: this is handled by `NewCipher`, which is not part of my code;
* overlap: I cannot do that without importing `unsafe`, which I'm hesitant to do.

crypto.go
jech: Done.

at-wat:
> This is not "rolling your own crypto". I'm just reimplementing a well-known, standard, boring piece of crypto machinery that has seen extensive peer review and that is very well understood.

`rolling your own crypto` includes reimplementing existing crypto. Keeping crypto implementation secured usually requires review of professionals of cryptology (I learned well but I'm not professional in it) and enough engineering resources for vulnerability management to address early warning information, which pion doesn't have and Go team has. (see https://golang.org/security)

jech:
I've revived this patch.  Updated benchmarks follow.  @Sean-Der please consider for merging, @davidzhao please provide encouragement ;-)

```
name                        old time/op    new time/op     delta
EncryptRTP-8                  1.58µs ± 8%     1.06µs ± 2%   -33.13%  (p=0.029 n=4+4)
EncryptRTPInPlace-8           1.54µs ± 3%     1.00µs ± 1%   -34.82%  (p=0.029 n=4+4)
DecryptRTP-8                  1.44µs ± 6%     0.67µs ± 2%   -53.80%  (p=0.029 n=4+4)
Write-8                       1.65µs ±10%     1.08µs ± 1%   -34.60%  (p=0.029 n=4+4)
WriteRTP-8                    1.52µs ± 2%     1.01µs ± 3%   -33.75%  (p=0.029 n=4+4)

name                        old speed      new speed       delta
EncryptRTP-8                71.1MB/s ± 8%  106.1MB/s ± 2%   +49.24%  (p=0.029 n=4+4)
EncryptRTPInPlace-8         72.9MB/s ± 3%  111.7MB/s ± 1%   +53.26%  (p=0.029 n=4+4)
DecryptRTP-8                19.4MB/s ± 6%   42.0MB/s ± 2%  +116.24%  (p=0.029 n=4+4)
Write-8                     68.2MB/s ± 9%  103.9MB/s ± 1%   +52.42%  (p=0.029 n=4+4)
WriteRTP-8                  73.6MB/s ± 2%  111.0MB/s ± 3%   +50.94%  (p=0.029 n=4+4)

name                        old alloc/op   new alloc/op    delta
EncryptRTP-8                    884B ± 0%       292B ± 0%   -66.97%  (p=0.029 n=4+4)
EncryptRTPInPlace-8             756B ± 0%       164B ± 0%   -78.31%  (p=0.029 n=4+4)
DecryptRTP-8                    780B ± 0%       188B ± 0%   -75.90%  (p=0.029 n=4+4)
Write-8                         916B ± 0%       324B ± 0%   -64.63%  (p=0.029 n=4+4)
WriteRTP-8                      804B ± 0%       212B ± 0%   -73.63%  (p=0.029 n=4+4)

name                        old allocs/op  new allocs/op   delta
EncryptRTP-8                    9.00 ± 0%       7.00 ± 0%   -22.22%  (p=0.029 n=4+4)
EncryptRTPInPlace-8             8.00 ± 0%       6.00 ± 0%   -25.00%  (p=0.029 n=4+4)
DecryptRTP-8                    9.00 ± 0%       7.00 ± 0%   -22.22%  (p=0.029 n=4+4)
Write-8                         9.00 ± 0%       7.00 ± 0%   -22.22%  (p=0.029 n=4+4)
WriteRTP-8                      8.00 ± 0%       6.00 ± 0%   -25.00%  (p=0.029 n=4+4)
```

Sean-Der:
@jech I am all for merging it! Sorry this didn't get done sooner.

Mind adding the error checks in the tests. Just `assert.NoError()` works great for me :)

jech:
Done.  (Sorry for the noise, but I don't know how to run ci-lint locally.)

Sean-Der:
This LGTM @jech! I can grab those last ones and will merge tonight

Sean-Der (approved):

adriancable:
@jech - this is amazing and drastically reduced my CPU usage (as per my post on #pion). Is there any chance you / @Sean-Der  can also revive PR #77 as well? The two go very well together.

PR #77, opened 2020-06-02
Avoid growbuffer

jech:
Allocation decreases by 70% for large packets.  CPU time remains unchanged for small packets, decreases slightly for large packets.
```

name            old time/op    new time/op    delta
Write14-4         2.73µs ± 0%    2.72µs ± 0%   -0.39%  (p=0.007 n=7+8)
Write140-4        3.16µs ± 0%    3.13µs ± 2%   -1.16%  (p=0.016 n=8+8)
Write1400-4       8.00µs ± 0%    7.58µs ± 0%   -5.24%  (p=0.000 n=8+8)
WriteRTP14-4      2.68µs ± 1%    2.67µs ± 0%   -0.36%  (p=0.035 n=8+7)
WriteRTP140-4     3.11µs ± 1%    3.08µs ± 1%   -0.98%  (p=0.005 n=7+8)
WriteRTP1400-4    7.97µs ± 0%    7.54µs ± 0%   -5.39%  (p=0.000 n=7+8)

name            old speed      new speed      delta
Write14-4       9.50MB/s ± 1%  9.55MB/s ± 0%   +0.51%  (p=0.007 n=8+8)
Write140-4      48.0MB/s ± 0%  48.6MB/s ± 2%   +1.18%  (p=0.016 n=8+8)
Write1400-4      176MB/s ± 0%   186MB/s ± 0%   +5.53%  (p=0.000 n=8+8)
WriteRTP14-4    9.71MB/s ± 1%  9.75MB/s ± 0%   +0.36%  (p=0.036 n=8+7)
WriteRTP140-4   48.9MB/s ± 1%  49.4MB/s ± 1%   +1.00%  (p=0.008 n=7+8)
WriteRTP1400-4   177MB/s ± 0%   187MB/s ± 0%   +5.70%  (p=0.000 n=7+8)

name            old alloc/op   new alloc/op   delta
Write14-4           708B ± 0%      660B ± 0%   -6.78%  (p=0.000 n=8+8)
Write140-4          836B ± 0%      660B ± 0%  -21.05%  (p=0.000 n=8+8)
Write1400-4       2.20kB ± 0%    0.66kB ± 0%  -69.95%  (p=0.000 n=8+8)
WriteRTP14-4        708B ± 0%      660B ± 0%   -6.78%  (p=0.000 n=8+8)
WriteRTP140-4       836B ± 0%      660B ± 0%  -21.05%  (p=0.000 n=8+8)
WriteRTP1400-4    2.20kB ± 0%    0.66kB ± 0%  -69.95%  (p=0.000 n=8+8)

name            old allocs/op  new allocs/op  delta
Write14-4           7.00 ± 0%      6.00 ± 0%  -14.29%  (p=0.000 n=8+8)
Write140-4          7.00 ± 0%      6.00 ± 0%  -14.29%  (p=0.000 n=8+8)
Write1400-4         7.00 ± 0%      6.00 ± 0%  -14.29%  (p=0.000 n=8+8)
WriteRTP14-4        7.00 ± 0%      6.00 ± 0%  -14.29%  (p=0.000 n=8+8)
WriteRTP140-4       7.00 ± 0%      6.00 ± 0%  -14.29%  (p=0.000 n=8+8)
WriteRTP1400-4      7.00 ± 0%      6.00 ± 0%  -14.29%  (p=0.000 n=8+8)
```

### DIFF
--- a/session_srtp.go
+++ b/session_srtp.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"sync"
 
 	"github.com/pion/logging"
 	"github.com/pion/rtp"
@@ -119,13 +120,29 @@ func (s *SessionSRTP) write(b []byte) (int, error) {
 	return s.writeRTP(&packet.Header, packet.Payload)
 }
 
+// bufferpool is a global pool of buffers used for encrypted packets in
+// writeRTP below.  Since it's global, buffers can be shared between
+// different sessions, which amortizes the cost of allocating the pool.
+var bufferpool = sync.Pool{
+	New: func() interface{} {
+		return make([]byte, 1460)
+	},
+}
+
 func (s *SessionSRTP) writeRTP(header *rtp.Header, payload []byte) (int, error) {
 	if _, ok := <-s.session.started; ok {
 		return 0, fmt.Errorf("started channel used incorrectly, should only be closed")
 	}
 
+	// encryptRTP will either return our buffer, or, if it is too
+	// small, allocate a new buffer itself.  In either case, it is
+	// safe to put the buffer back into the pool, but only after
+	// nextConn.Write has returned.
+	buf := bufferpool.Get()
+	defer bufferpool.Put(buf)
+
 	s.session.localContextMutex.Lock()
-	encrypted, err := s.localContext.encryptRTP(nil, header, payload)
+	encrypted, err := s.localContext.encryptRTP(buf.([]byte), header, payload)
 	s.session.localContextMutex.Unlock()
 
 	if err != nil {
--- a/srtp_test.go
+++ b/srtp_test.go
@@ -359,6 +359,7 @@ func BenchmarkEncryptRTP(b *testing.B) {
 		b.Fatal(err)
 	}
 
+	b.SetBytes(int64(len(pktRaw)))
 	b.ResetTimer()
 
 	for i := 0; i < b.N; i++ {
@@ -383,6 +384,7 @@ func BenchmarkEncryptRTPInPlace(b *testing.B) {
 
 	buf := make([]byte, 0, len(pktRaw)+10)
 
+	b.SetBytes(int64(len(pktRaw)))
 	b.ResetTimer()
 
 	for i := 0; i < b.N; i++ {
@@ -414,6 +416,7 @@ func BenchmarkDecryptRTP(b *testing.B) {
 		b.Fatal(err)
 	}
 
+	b.SetBytes(int64(len(encryptedRaw)))
 	b.ResetTimer()
 
 	for i := 0; i < b.N; i++ {
--- a/srtp_test.go
+++ b/srtp_test.go
@@ -347,13 +347,13 @@ func TestRTPReplayProtection(t *testing.T) {
 	}
 }
 
-func BenchmarkEncryptRTP(b *testing.B) {
+func benchmarkEncryptRTP(b *testing.B, size int) {
 	encryptContext, err := buildTestContext()
 	if err != nil {
 		b.Fatal(err)
 	}
 
-	pkt := &rtp.Packet{Payload: make([]byte, 100)}
+	pkt := &rtp.Packet{Payload: make([]byte, size)}
 	pktRaw, err := pkt.Marshal()
 	if err != nil {
 		b.Fatal(err)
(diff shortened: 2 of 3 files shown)


Sean-Der:
Amazing! @jech is this worth doing for RTCP as well?

Happy to tag/merge whenever you are happy with it.

jech:
Unless I'm mistaken, SRTCP uses `allocateIfMismatch` instead of `growBuffer`, and I'm not quite sure what this code is doing.

jech:
Sean, I suggest merging the current version, I'm not comfortable using manual memory allocation in the SRTCP writer until I fully grok it.

Sean-Der:
@jech one thing I realized is this might be confusing behavior for non-`pion/webrtc` users. This means that the `nextConn` must not keep the packet.

I think this is another one we need to add behind a flag, sorry for the annoyance :(

If you are really looking for a performance improvement have you looked at `AES-GCM`, I think that will give you a much bigger payoff! We had a Pion user that implemented it, but then never upstreamed. He posted his results [here](https://bugs.chromium.org/p/chromium/issues/detail?id=713701#c20)

jech:
> This means that the nextConn must not keep the packet.

Yes, that's the direction I'm moving in.

In https://github.com/pion/webrtc/issues/1227, you mentioned that there is inefficiency in Pion because it's structured as a set of modules.  I don't necessarily agree this is a problem, as long as there are clearly defined protocols between the modules.

The protocol I'm assuming is that all `Write`-like functions don't retain the buffer (they copy the data), and that all the `Read`-like functions copy internal data rather than returning it.  This is a simple protocol that is consistent with what the Go stdlib does.  Pion mostly obeys this protocol.  The one exception I've found is `samplebuilder`, as outlined in https://github.com/pion/webrtc/issues/1214.

Now if there is no protocol between the parts of Pion, then you're right, there's nothing that can be done to avoid massive amounts of overhead.

jech:
I am wondering whether you could reconsider this issue for v3.

It is a general rule in Go that the `Write` method of an `io.Writer` does not retain the buffer.  It is legal to write
```
for {
    n, _ := r.Read(buf)
    w.Write(buf[:n])
}
```
With no such guarantee, the caller must allocate a buffer on each iteration, which causes pressure on the memory subsystem and harms cache locality.

This disciplne is generally obeyed in Pion.  The only place I've found where it is violated is the sample builder, which I'm planning to fix at some point.

OrlandoCo:
@jech @Sean-Der This looks like a good improvement to reduce allocations, but to properly do it the sync.Pool should store/return a pointer to the slice:

```go
	var bufferpool = sync.Pool{
		New: func() interface{} {
			b:= make([]byte, 1460)
			return &b
		},
	}
```

Also to measure the impact of the sync.Pool I add it to the bench, a good thing about it is that even that pool is thread safe it does no use any lock, so there is not an issue to put a global lock to the writer:

```go
	for i := 0; i < b.N; i++ {
		src:=bufferpool.Get().(*[]byte)
		_, err := encryptContext.EncryptRTP(*src, pktRaw, nil)
		if err != nil {
			b.Fatal(err)
		}
		bufferpool.Put(src)
	}
```
This are the results of the bechs

```bash
## No sync pool
BenchmarkEncryptRTP14
BenchmarkEncryptRTP14-6                   784328              1309 ns/op             700 B/op          7 allocs/op
BenchmarkEncryptRTP140
BenchmarkEncryptRTP140-6                  771495              1544 ns/op             828 B/op          7 allocs/op
BenchmarkEncryptRTP1400
BenchmarkEncryptRTP1400-6                 298354              4261 ns/op            2188 B/op          7 allocs/op
##  Sync pool with []byte
BenchmarkEncryptRTPInPlace14
BenchmarkEncryptRTPInPlace14-6           1000000              1368 ns/op             676 B/op          7 allocs/op
BenchmarkEncryptRTPInPlace140
BenchmarkEncryptRTPInPlace140-6          1000000              1570 ns/op             676 B/op          7 allocs/op
BenchmarkEncryptRTPInPlace1400
BenchmarkEncryptRTPInPlace1400-6          315472              3969 ns/op             676 B/op          7 allocs/op
## Sync pool with *[]byte
BenchmarkEncryptRTPInPlace14
BenchmarkEncryptRTPInPlace14-6            915198              1288 ns/op             652 B/op          6 allocs/op
BenchmarkEncryptRTPInPlace140
BenchmarkEncryptRTPInPlace140-6          1000000              1498 ns/op             652 B/op          6 allocs/op
BenchmarkEncryptRTPInPlace1400
BenchmarkEncryptRTPInPlace1400-6          295365              3884 ns/op             652 B/op          6 allocs/op
```

jech:
@OrlandoCo, you are allocating a new `interface{}` buffer for each pool allocation.  Your code should look like this:
```
isrc := bufferpool.Get()
src := isrc.(*[]byte)
...
bufferpool.Put(isrc)
```
Note how I'm keeping the original `interface{}` wrapper rather than allocating a new one at `Put` time.

If you do this, I'm pretty sure that the `[]byte` version will have almost exactly the same performance as the `*[]byte` one.

OrlandoCo:
@jech oh yeah I usually forgot that interface conversion allocs sometimes, but anyways whatever path we follow, is something stopping you from merging this?

jech:
I don't have commit rights to Pion.

jech:
Obsoleted by #197.

PR #82, opened 2020-07-09
Avoid ROC update by unauthorized packets

at-wat:
ROC must be updated after auth tag validation.
Otherwise, ROC can be manipulated by attacker.


at-wat:
Adding a new tag v1.4.0

Issue #85, opened 2020-07-14
Add AES-GCM Support


Sean-Der:
I am working on a benchmark to show `RTSP -> WebRTC` fanout performance. I want to land this first before getting numbers :)

This is the difference when using libsrtp, so we should see a pretty big improvement for Pion as well! This is due to be able to use hardware acceleration (Auth+Encrypt) can all be done in one step now.

```
-- aes_cm_128_hmac_sha1_32
./a.out  1.83s user 0.01s system 99% cpu 1.843 total
./a.out  1.84s user 0.01s system 99% cpu 1.856 total
./a.out  1.83s user 0.00s system 99% cpu 1.837 total
./a.out  1.82s user 0.01s system 99% cpu 1.831 total
./a.out  2.04s user 0.01s system 99% cpu 2.065 total
./a.out  1.82s user 0.01s system 99% cpu 1.834 total
./a.out  1.84s user 0.01s system 99% cpu 1.857 total
./a.out  1.83s user 0.00s system 99% cpu 1.843 total
./a.out  1.85s user 0.01s system 99% cpu 1.860 total
./a.out  1.83s user 0.01s system 99% cpu 1.839 total

-- aes_gcm_128_16
./a.out  0.76s user 0.00s system 99% cpu 0.772 total
./a.out  0.81s user 0.00s system 99% cpu 0.814 total
./a.out  0.75s user 0.00s system 99% cpu 0.762 total
./a.out  0.75s user 0.01s system 99% cpu 0.760 total
./a.out  0.76s user 0.00s system 99% cpu 0.765 total
./a.out  0.76s user 0.00s system 99% cpu 0.765 total
./a.out  0.79s user 0.01s system 99% cpu 0.803 total
./a.out  0.77s user 0.01s system 99% cpu 0.780 total
./a.out  0.75s user 0.00s system 99% cpu 0.762 total
./a.out  0.79s user 0.01s system 99% cpu 0.796 total
```

Sean-Der:
https://tools.ietf.org/html/rfc7714 is the RFC for this

at-wat:
Latest shares of WebRTC SRTP cipher
[...]

PR #87, opened 2020-07-16
Generalize ProtectionProfiles

Sean-Der:
This removes a lot of profile specific constants and insteads
put them behind the profile. Key, Salt and TagLen need to be
determined by the profile.

Relates to #85


Sean-Der:
I didn't change any APIs, just moved all the AES-CM stuff into their own files! Wanted to get this refactoring/changes in its own commit to make debugging easier if anything happens.

When this gets approved will open up AES-GCM in next commit!

PR #88, opened 2020-07-19
Implement AEAD_AES_128_GCM

### LINKED ISSUE #85: Add AES-GCM Support

Sean-Der:
Resolves #85


key_derivation.go
```diff
@@ -5,56 +5,28 @@ import (
...
 	block, err := aes.NewCipher(masterKey)
 	if err != nil {
 		return nil, err
 	}
 
-	block.Encrypt(sessionSalt, sessionSalt)
-	return sessionSalt[0:len(masterSalt)], nil
+	block.Encrypt(out, out)
+	return out[:outLen], nil
 }
```
at-wat, line 30:
```go
func aesCmKeyDerivation(label byte, masterKey, masterSalt []byte, indexOverKdr int, outLen int) ([]byte, error) {
  // https://tools.ietf.org/html/rfc3711#appendix-B.3
  // The input block for AES-CM is generated by exclusive-oring the master salt with the
  // concatenation of the encryption key label 0x00 with (index DIV kdr),
  // - index is 'rollover count' and DIV is 'divided by'
  
  nMasterKey := len(masterKey)
  nMasterSalt := len(masterSalt)

  prfIn := make([]byte, nMasterKey)
  copy(prfIn[:nMasterSalt], masterSalt)

  prfIn[7] ^= label
  if indexOverKdr != 0 {
    // 24-bit "index DIV kdr" must be xored to prf input.
    return nil, errors.New("indexOverKdr > 0 is not supported yet")
  }

  //The resulting value is then AES encrypted using the master key to get the cipher key.
  block, err := aes.NewCipher(masterKey)
  if err != nil {
    return nil, err
  }

  out := make([]byte, ((outLen+nMasterKey)/nMasterKey)*nMasterKey)
  var i uint16
  for n := 0; n < outLen; n += nMasterKey {
    binary.BigEndian.PutUint16(prfIn[nMasterKey-2:], i)
    block.Encrypt(out[n:n+nMasterKey], prfIn)
    i++
  }
  return out[:outLen], nil
}
```
unifies `generateSessionAuthTag` into `aesCmKeyDerivation`.

srtp_cipher_aead_aes_gcm_test.go
```diff
@@ -0,0 +1,85 @@
+package srtp
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSrtpCipherAedAesGcm(t *testing.T) {
```
at-wat, line 9: Failing on CI.

srtp_cipher_aead_aes_gcm.go
```diff
@@ -0,0 +1,171 @@
...
+	"encoding/binary"
+
+	"github.com/pion/rtp"
+)
+
+const (
+	rtcpEncryptionFlag = 0x80
+)
+
+type srtpCipherAedAesGcm struct {
```
at-wat, line 15:
```suggestion
type srtpCipherAeadAesGcm struct {
```

srtp_cipher_aead_aes_gcm.go
```diff
@@ -0,0 +1,171 @@
...
+	srtpCipher, srtcpCipher cipher.AEAD
+
+	srtpSessionSalt, srtcpSessionSalt []byte
+}
+
+func newSrtpCipherAedAesGcm(masterKey, masterSalt []byte) (*srtpCipherAedAesGcm, error) {
```
at-wat, line 21:
```suggestion
func newSrtpCipherAeadAesGcm(masterKey, masterSalt []byte) (*srtpCipherAedAesGcm, error) {
```

srtp_cipher_aead_aes_gcm.go
Sean-Der: Oof fixed, nice catch :)

srtp_cipher_aead_aes_gcm.go
Sean-Der: Fixed!

context.go
```diff
@@ -75,16 +75,23 @@ func CreateContext(masterKey, masterSalt []byte, profile ProtectionProfile, opts
...
 		srtcpSSRCStates: map[uint32]*srtcpSSRCState{},
 	}
+
+	switch profile {
+	case ProtectionProfileAeadAes128Gcm:
+		c.cipher, err = newSrtpCipherAedAesGcm(masterKey, masterSalt)
+	case ProtectionProfileAes128CmHmacSha1_80:
+		c.cipher, err = newSrtpCipherAesCmHmacSha1(masterKey, masterSalt)
+	default:
+		return nil, fmt.Errorf("No such SRTP Profile %#v", profile)
```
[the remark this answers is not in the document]
Sean-Der: Fixed]

srtp_cipher_aead_aes_gcm_test.go
Sean-Der: I had messed up RTCP, used `srtpCipher` instead of `srtcpCipher` when refactoring :(

key_derivation.go
Sean-Der: Wow this is an amazing cleanup! Thanks for digging so deep into this PR!

key_derivation.go
```diff
@@ -5,56 +5,28 @@ import (
...
 
 func generateSessionAuthTag(label byte, masterKey, masterSalt []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
Sean-Der: Done! Really amazing refactor/cleanup. This code has been like this for years, and never thought of doing this.

srtp_cipher_aead_aes_gcm.go
```diff
@@ -0,0 +1,171 @@
...
+	rtcpEncryptionFlag = 0x80
+)
+
+type srtpCipherAeadAesGcm struct {
+	srtpCipher, srtcpCipher cipher.AEAD
+
+	srtpSessionSalt, srtcpSessionSalt []byte
+}
+
+func newSrtpCipherAedAesGcm(masterKey, masterSalt []byte) (*srtpCipherAeadAesGcm, error) {
```
at-wat, line 21:
```suggestion
func newSrtpCipherAeadAesGcm(masterKey, masterSalt []byte) (*srtpCipherAeadAesGcm, error) {
```

srtp_cipher_aes_cm_hmac_sha1.go
```diff
@@ -23,34 +23,36 @@ type srtpCipherAesCmHmacSha1 struct {
 	srtcpBlock       cipher.Block
 }
 
+const srtpCipherAesCmHmacSha1AuthKeySize = 20
```
at-wat, line 26: I think this is better to be `ProtectionProfile.authKeyLen()` instead of const.

srtp_cipher_aead_aes_gcm.go
Sean-Der: Fixed!

srtp_cipher_aes_cm_hmac_sha1.go
Sean-Der: AES-GCM doesn't have an authKeyLen, ok if I add `authKeyLen` to `srtpCipherAesCmHmacSha1` ?

at-wat: For AES-GCM case, returning 0 would be just fine.

Sean-Der: Done!

srtp_cipher.go
```diff
@@ -6,6 +6,7 @@ import "github.com/pion/rtp"
 // of the SRTP Specific ciphers
 type srtpCipher interface {
 	authTagLen() int
```
at-wat, line 8: https://tools.ietf.org/html/rfc3711#section-5.2
>   HMAC-SHA1, as defined in Section 4.2.1, SHALL be the default message
>   authentication code.  The default session authentication key-length
>   (n_a) SHALL be 160 bits, the default authentication tag length
>   (n_tag) SHALL be 80 bits, and the SRTP_PREFIX_LENGTH SHALL be zero
>   for HMAC-SHA1.  In addition, for SRTCP, the pre-defined HMAC-SHA1
>   MUST NOT be applied with a value of n_tag, nor n_a, that are smaller
>   than these defaults.  For SRTP, smaller values are NOT RECOMMENDED,
>   but MAY be used after careful consideration of the issues in Section
>   7.5 and 9.5.

In my understanding, RFC only specifies default value.
User can select shorter/longer auth tag len (e.g. 32-bit tag is pre-defined) independent from the algorithm defiintion on HMAC-SHA1. (On GCM, it must be always 16 bytes.)

Sean-Der: Yea! So `srtp_cipher_aes_cm_hmac_sha1` should be a little more general for `ProtectionProfileAes128CmHmacSha1_32`. I can make that change, but I don't think it is worth it (for WebRTC users)

FireFox dropped support (might not be done yet) for `SRTP_AES128_CM_HMAC_SHA1_32` and I think everyone else followed.

at-wat: For WebRTC case, yes. But for more general IoT usecases, I think 32-bit tag is still worth. (https://tools.ietf.org/html/rfc3711#section-7.5)

Sean-Der: :+1: Yea I am totally open to it being added! I don't plan on doing the work right now though (unless we have a user that is blocked)

Sean-Der: @at-wat Do you want me to implement that in this PR? Or can I create a follow up issue?

at-wat: I think it's fine to be a follow up issue!

key_derivation_test.go
```diff
@@ -13,24 +15,32 @@ func TestValidSessionKeys(t *testing.T) {
...
 	if err != nil {
 		t.Errorf("generateSessionAuthTag failed: %v", err)
 	} else if !bytes.Equal(sessionAuthTag, expectedSessionAuthTag) {
 		t.Errorf("Session Auth Tag % 02x does not match expected % 02x", sessionAuthTag, expectedSessionAuthTag)
 	}
 }
+
+func TestIndexOverKDR(t *testing.T) {
+	_, err := aesCmKeyDerivation(labelSRTPAuthenticationTag, []byte{}, []byte{}, 1, 0)
+	assert.Error(t, err)
```
at-wat, line 45: It's not expected to return error, but currently not implemented yet. Better to add a comment.

Sean-Der: :+1: added comment.

Issue #9, opened 2019-02-17
`EncryptRTP` API change


kixelated:
**current:**
```
EncryptRTP(dst []byte, plaintext []byte, header *rtp.Header) ([]byte, error)
```

**proposed:**
```
EncryptRTP(dst []byte, packet *rtp.Packet) ([]byte, error)
```

This is a simpler API when `*rtp.Packet` is utilized.

It subsequently allows for some small performance optimizations:
1. Avoids an extra Unmarshal for callers who use `rtp.Packet` (ex: `WriteRTP`).
2. Avoids a memcpy of the payload prior to encryption.
3. Removes the implicit case that `cap(dst) >= len(plaintext) + 10` in order to avoid an allocation.

This API is more cumbersome for callers that deal with byte slices, as they would have to `Unmarshal` prior to encryption. However, there is no performance impact.

kixelated:
ditto for the `WriteRTP` and `ReadRTP` methods. Should combine `Header, []byte` into `Packet` everywhere.

sirzooro:
I am against this. In my case RTP/RTCP packets are generated by other parts of the system, and I use `pion/srtp` to encrypt generated packets. The same case is for receiving, I need to decrypt them to `[]byte` and pass elsewhere for processing.

PR #113, opened 2020-11-28
Migrate to use ConnCtx

at-wat:
Use Conn controlled by context.Context.
If you want to use net.Conn as a underlying connection, wrapping it by github.com/pion/transport/ctxconn.ConnCtx would be the easiest way.


at-wat:
```
failed to initialize build cache at /.cache/go-build: mkdir /.cache/go-build: permission denied
Error: Process completed with exit code 1.
```
on i386.
I think CI cache key should contain target arch since the file owner is different.  (x86_64: builder, i386: root (in docker))

go.mod
```diff
@@ -7,6 +7,6 @@ require (
 	github.com/pion/logging v0.2.2
 	github.com/pion/rtcp v1.2.4
 	github.com/pion/rtp v1.6.1
-	github.com/pion/transport v0.10.1
+	github.com/pion/transport v0.10.2-0.20201128143537-2be49f0a5e9c
```
at-wat, line 10: Update after releasing https://github.com/pion/transport/pull/108

Sean-Der (dismissed):

Sean-Der:
@at-wat mind starting `/v2` for this? Since we are breaking public APIs

at-wat:
@Sean-Der I think it's better to consider https://github.com/pion/srtp/issues/9 before releasing v2.0.0.
Let's start v2.0.0-rc0.

Sean-Der:
Works for me! I am somewhat hesitant to do more changes, I really want to release pion/webrtc v3. We have people building with v2 and it’s a decent support burden.

go.mod
at-wat: done

Sean-Der (approved):

PR #126, opened 2020-12-18
Convert packetIO buffer to an interface

OrlandoCo:
#### Description
PacketIO buffer interface will allow to use custom buffer for rtp/rtcp streams

#### Reference 
https://github.com/pion/transport/pull/121


jech:
What's the performance hit?

OrlandoCo:
> What's the performance hit?

@jech It's negligible

```bash
# Interface buffer
BenchmarkWriteRTP-6   	  807013	      1424 ns/op	     820 B/op	       8 allocs/op
# None interface
BenchmarkWriteRTP-6   	  790890	      1411 ns/op	     820 B/op	       8 allocs/op
```

session.go
```diff
@@ -40,6 +42,7 @@ type session struct {
 type Config struct {
 	Keys          SessionKeys
 	Profile       ProtectionProfile
+	BufferFactory func(packetType packetio.BufferPacketType, ssrc uint32) packetio.Buffer
```
jech, line 45: BufferFactory should return an io.ReadWriteCloser.  The calling code should dynamically check whether the relevant methods are available, and ignore them if they do not.

stream_srtcp.go
```diff
@@ -100,9 +100,13 @@ func (r *ReadStreamSRTCP) init(child streamSession, ssrc uint32) error {
 	r.isInited = true
 	r.isClosed = make(chan bool)
 
-	// Create a buffer and limit it to 100KB
-	r.buffer = packetio.NewBuffer()
-	r.buffer.SetLimitSize(srtcpBufferSize)
+	if r.session.bufferFactory != nil {
+		r.buffer = r.session.bufferFactory(packetio.RTCPBufferPacket, ssrc)
+	} else {
+		// Create a buffer and limit it to 100KB
```
jech, line 106: This comment merely paraphrases the code, and risks becoming out of sync if we change srtcpBufferSize.

OrlandoCo: It's from original code, not mine

session.go
OrlandoCo: For RTCP/RTP paths is safe to stick with io.ReadWriteCloser, since is believe is the only useful case for the buffer factory I agree to make this change

OrlandoCo: Done!

jech:
Clean API, includes test, and we have a convincing benchmark.  I like it, and strongly support.

I have one minor nit and one question I'd like you to consider: 

  * I'd prefer the struct to be called `Buffer` (exported), `NewBuffer` to return `*Buffer` (as it used to), and to avoid creating a new interface;
  * should the `BufferFactory` take some extra parameters beyond the SSID?  Perhaps the PeerConnection?  The RTPSender or Receiver?

OrlandoCo:
@jech 

> * I'd prefer the struct to be called `Buffer` (exported), `NewBuffer` to return `*Buffer` (as it used to), and to avoid creating a new interface;

I updated it earlier, now Buffer is the struct as always and the interface is `io.ReadeWriteCloser` so no new interface was introduced

> * should the `BufferFactory` take some extra parameters beyond the SSID?  Perhaps the PeerConnection?  The RTPSender or Receiver?

I personally would like to get a RTCPWriter /RTPWriter, so I could include the congestion control directly in the buffer, that would clean a lot of logic in the SFU and give far better performance than interceptors

jech:
>     * I'd prefer the struct to be called Buffer (exported), NewBuffer to return
>       *Buffer (as it used to), and to avoid creating a new interface;

> I updated it earlier, now Buffer is the struct as always and the interface is
> io.ReadeWriteCloser so no new interface was introduced

Excellent.  Sorry for the confusion, but I find Github's interface confusing.

>     * should the BufferFactory take some extra parameters beyond the SSID?
>       Perhaps the PeerConnection? The RTPSender or Receiver?

> I personally would like to get a RTCPWriter /RTPWriter

Yeah, something like that.

But I'm not sure how that would work.  You get your receiver report, and
now you need to find the data structure to update with the packet loss
data you just received (this could be the Track, this could be the
RTPsender, or it could be some data structure that encapsulates the Track,
as in the case of Galène).  I'd expect the BufferFactory to stick
a reference to the right data structure into your ReadWriteCloser, but for
that, we need the BufferFactory to have access to the relevant data.

jech:
I see the following ways forward:

1. Merge the Buffer changes, but name the factory method FactoryInternal.  Replace it later with a proper factory method, when Orlando has grokked the fullness.
2. Merge the Buffer changes, but do not merge the factory method.  Add a method `ReplaceBuffer` which allows replacing the buffer associated with a given transceiver.
3. Merge the Buffer changes, do not merge the factory method.  Have ion and Galène use a forked version of Pion until we can agree on a Factory method.

Ihave a preference for (3) — it's always easy to add a new method, it's always problematic to change one.

OrlandoCo:
> I see the following ways forward:
> 
> 1. Merge the Buffer changes, but name the factory method FactoryInternal.  Replace it later with a proper factory method, when Orlando has grokked the fullness.
> 2. Merge the Buffer changes, but do not merge the factory method.  Add a method `ReplaceBuffer` which allows replacing the buffer associated with a given transceiver.
> 3. Merge the Buffer changes, do not merge the factory method.  Have ion and Galène use a forked version of Pion until we can agree on a Factory method.
> 
> Ihave a preference for (3) — it's always easy to add a new method, it's always problematic to change one.

I have been updating ion-sfu, and I think the best approach is to share something like `interceptor.StreamInfo` the issue is that we have all the info to do that but the modules makes hard to share the info. So I dont have enough info to relate a SSRC to a peer and a codec because it takes some time to fire the track event, is a shame because that way I could get the real arrive time etc.

Going to keep thinking the best approach.

tarrencev (approved):

PR #129, opened 2020-12-22
Added read timeouts

Threadnaught:
#### Added timeouts to ReadStreamSRTCP and ReadStreamSRTP

As part of a project that I'm currently working on I needed webrtc to be able to timeout during a channel read. I discussed how to do this with Sean and he suggested adding a function here that calls through to the underlying `buffer. SetReadDeadline`. I have another change ready to add to the WebRTC once this is approved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Novel Corpse](https://github.com/NovelCorpse)
 * [Jerko Steiner](https://github.com/jeremija)
 * [Juliusz Chroboczek](https://github.com/jech)
+* [Threadnaught](https://github.com/Threadnaught/)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/stream_srtcp.go
+++ b/stream_srtcp.go
@@ -3,6 +3,7 @@ package srtp
 import (
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/pion/rtcp"
 	"github.com/pion/transport/packetio"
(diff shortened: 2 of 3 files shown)


mission-liao:
@Threadnaught Hi, appreciated for this PR, but I believe this part is done in this commit https://github.com/pion/srtp/commit/9fa5210f96035bb6089b906c5736e458d61de699.

The next step is to expose these timeout methods to `pion/webrtc`, I've reopne this [issue](https://github.com/pion/webrtc/issues/1553) but haven't started the work, please feel free to take over it.

Threadnaught:
@mission-liao I'm not convinced that `SessionSRT(C)P` is the best place to put that. `session` is unexported in `ReadStreamSRT(C)P`, and `setReadDeadline` on `SessionSRT(C)P` is also unexported in the implementation that you reference. They need to be accessible from `webrtc`, which is a different package. I guess I could rewrite this change so that it calls those functions, but wouldn't it be more direct to just call it against the `buffer` in `ReadStreamSRT(C)P`?

Threadnaught:
Oh by the way, how do I run the tests?

mission-liao:
@Threadnaught Hi, I'm not sure if I understand your statement correctly, please feel free to correct me.

- The last PR aims to export `SetReadDeadline` from `ReadStreamSR(C)T`, and yes, it simply calls `buffer`'s `SetReadDealine` method. refer [here](https://github.com/pion/srtp/blob/9fa5210f96035bb6089b906c5736e458d61de699/stream_srtp.go#L92)
- Yes, `pion/webrtc` can't access `buffer` member in `ReadStreamSR(C)T`, that's why we need to export its timeout method by adding `SetReadDeadline` to `ReadStreamSR(C)T`. And then submit another PR to `pion/webrtc` to  use it.
- To run unittest, simply run this command `go test -v ./...`.

Threadnaught:
I somehow managed to miss the changes you made to `ReadStreamSRT(C)P`, when I skimmed that PR, sorry. I'll see if my changes to webrtc need any adaptation to work with your version of the code.

mission-liao:
Happy committing

PR #133, opened 2021-01-19
Fix panic when calling SetReadDeadline

mission-liao:
#### Description
The io.ReadWriteCloser doesn't include this method, should make sure it
exists before calling it.

After integrated the `webrtc.SettingEngine.BufferFactory`, this issue arises.


mission-liao:
@OrlandoCo I found this issue when adapting `SettingEngine.BufferFactory` and `ion-sfu/pkg/buffer` in my project, just propose a possible fix. One concern to me: should we silently return `nil` when the buffer doesn't support deadline capability.

OrlandoCo:
@mission-liao I should say that yeah, since with custom buffer you can set whatever deadline method you want,  and not enforce to use one.

Sean-Der (approved):

Sean-Der:
Sorry this took so long! Feel free to tag @mission-liao :)

PR #137, opened 2021-03-06
Fix SRTCP decrypt for AEAD_AES_GCM

at-wat:
- Fix AEAD auth tag handling
- Add test cases for AEAD_AES_GCM

See https://tools.ietf.org/html/rfc7714#section-16.1.2 about the auth tag location in AEAD_AES_GCM mode.

### DIFF
--- a/srtcp_test.go
+++ b/srtcp_test.go
@@ -10,46 +10,170 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func rtcpTestMasterKey() []byte {
-	return []byte{0xfd, 0xa6, 0x25, 0x95, 0xd7, 0xf6, 0x92, 0x6f, 0x7d, 0x9c, 0x02, 0x4c, 0xc9, 0x20, 0x9f, 0x34}
-}
-
-func rtcpTestMasterSalt() []byte {
-	return []byte{0xa9, 0x65, 0x19, 0x85, 0x54, 0x0b, 0x47, 0xbe, 0x2f, 0x27, 0xa8, 0xb8, 0x81, 0x23}
-}
-
-type rtcpTestCase struct {
+type rtcpTestPacket struct {
 	ssrc      uint32
 	index     uint32
+	pktType   rtcp.PacketType
 	encrypted []byte
 	decrypted []byte
 }
 
-func rtcpTestCases() []rtcpTestCase {
-	return []rtcpTestCase{
-		{
-			ssrc:      0x66ef91ff,
-			index:     0,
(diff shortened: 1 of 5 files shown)


Sean-Der:
@at-wat Could we add a conditional to the test for `isAEAD`? We could even keep it private.

This is just adding more test, or is more code coming? thanks!

srtcp_test.go
```diff
@@ -63,182 +187,228 @@ func TestRTCPLifecycle(t *testing.T) {
 	for name, option := range options {
 		option := option
 		t.Run(name, func(t *testing.T) {
+			for caseName, testCase := range rtcpTestCases() {
+				testCase := testCase
+				t.Run(caseName, func(t *testing.T) {
+					if testCase.algo == ProtectionProfileAeadAes128Gcm {
+						t.Skip("FIXME: DecryptRTCP(nil, input, nil) for ProtectionProfileAeadAes128Gcm changes input data")
```
at-wat, line 194: It should be fixed but out of the scope of this PR.

at-wat:
> Could we add a conditional to the test for `isAEAD`? We could even keep it private.

@Sean-Der Do you mean to add `isAEAD() bool` to `ProtectionProfile`? I'm not sure how it will be used.

Sean-Der:
I don't understand the need for `aeadAuthTagLen` vs `authTagLen`. Is there a case where both would be non-zero?

In the test why not do

```
                if cipherSuiteIsAEAD > 0 {
                    authTagPos := len(rtcpPacket) - authTagLen - srtcpIndexSize
                    copy(rtcpPacket[authTagPos:authTagPos+authTagLen], make([]byte, authTagLen))
                } else {
                    copy(rtcpPacket[len(rtcpPacket)-authTagLen:], make([]byte, authTagLen))
                }
```

Sean-Der:
But I would love a fix for this! The more tests the better as well. If you are able to submit a fix that would be great. I can take this tomorrow also now that we have tests.

at-wat:
RFC specifies auth tag and AEAD auth tag separately:
https://tools.ietf.org/html/rfc7714#section-14.2

> ```
>     SRTP_AEAD_AES_128_GCM
>          cipher:                 AES_128_GCM
>          cipher_key_length:      128 bits
>          cipher_salt_length:     96 bits
>          aead_auth_tag_length:   16 octets
>          auth_function:          NULL
>          auth_key_length:        N/A
>          auth_tag_length:        N/A
>          maximum lifetime:       at most 2^31 SRTCP packets and
>                                    at most 2^48 SRTP packets
> ```

as their location in the binary stream is different.

The confusion of `authTagLen` and `aeadAuthTagLen` caused a bug at
https://github.com/pion/srtp/blob/369037163aea2834a9f8d69e44d7032b1a0f6cee/srtcp.go#L12-L20

Sean-Der (approved):

PR #138, opened 2021-03-06
Fix destination buffer handling of AEAD cipher

at-wat:
- Fix not to change input buffer on decrypt
- Support in-place encrypt/decrypt

branched from #137

### DIFF
--- a/srtcp_test.go
+++ b/srtcp_test.go
@@ -190,9 +190,6 @@ func TestRTCPLifecycle(t *testing.T) {
 			for caseName, testCase := range rtcpTestCases() {
 				testCase := testCase
 				t.Run(caseName, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


Sean-Der:
Fantastic, nice work @at-wat!

I don't we should have two fields for tags. Right now it looks like we could have two tags with different lengths. 

Why can't we just have something denote `This cipher is AEAD`? Because what we are trying to communicate is the auth tag is  in a different location.

at-wat:
@Sean-Der I think confusion of authKey and aeadAuthKey actually caused a bug in srtcp decode function. I believe it should be separated to keep algorithm consistent and simple to avoid such bug.
Please see https://github.com/pion/srtp/pull/137#issuecomment-791885988 for details.

Sean-Der:
That's a great point. Also missed how the RFC had them different.

Merging! Will also tag new version of webrtc

PR #148, opened 2021-06-26
Session's contexts can now be updated

Antonito:
This PR adds a `UpdateContext` to the `session` object, which allow to change the context of an already running SRTP / SRTCP session.

This is needed for DTLS restart, as mentioned in pion/webrtc#1636.

Apart from the `UpdateContext`, this PR changes two things:

- `remoteContext` is now a `atomic.Value`: it previously had no locking mechanism, and none is required unless a context update is triggered.
I did not benchmark anything but I assume an atomic Value will be far more efficient than a `sync.RwLock` in this specific case.

- `session.start` no longer keeps its `localContext` if the creation of `remoteContext` fails


session.go
```diff
@@ -106,17 +108,19 @@ func (s *session) close() error {
 }
 
 func (s *session) start(localMasterKey, localMasterSalt, remoteMasterKey, remoteMasterSalt []byte, profile ProtectionProfile, child streamSession) error {
-	var err error
-	s.localContext, err = CreateContext(localMasterKey, localMasterSalt, profile, s.localOptions...)
+	localContext, err := CreateContext(localMasterKey, localMasterSalt, profile, s.localOptions...)
```
Sean-Der, line 111: can we call `s.UpdateContext`? Just brings down the duplicated code (hopefully)

Sean-Der:
Do we also need to update the `rolloverCounter` and `lastSequenceNumber` ? I will find the code in libwebrtc

session.go
Antonito: You're absolutely right, plus it removes the need for the extra unit test

Antonito:
> Do we also need to update the `rolloverCounter` and `lastSequenceNumber` ? I will find the code in libwebrtc

I decided to take a deep look at RFC 3711, and I don' t think the `rolloverCounter` should be updated:

```
*  a 32-bit unsigned rollover counter (ROC), which records how many
      times the 16-bit RTP sequence number has been reset to zero after
      passing through 65,535.  Unlike the sequence number (SEQ), which
      SRTP extracts from the RTP packet header, the ROC is maintained by
      SRTP as described in [Section 3.3.1](https://datatracker.ietf.org/doc/html/rfc3711#section-3.3.1).

      We define the index of the SRTP packet corresponding to a given
      ROC and RTP sequence number to be the 48-bit quantity

            i = 2^16 * ROC + SEQ.
```

About the `lastSequenceNumber`, I don't think we should update it either:

```
*  for the receiver only, a 16-bit sequence number s_l, which can be
      thought of as the highest received RTP sequence number (see
      [Section 3.3.1](https://datatracker.ietf.org/doc/html/rfc3711#section-3.3.1) for its handling), which SHOULD be authenticated
      since message authentication is RECOMMENDED,
```

There's a concept of re-keying (for instance if a key expire), during which both `lastSequenceNumber` and `ROC` have a role to play, but that's different from what we do here – we use the same key.

```
Note that if the master key is to be shared between SRTP streams
   within the same RTP session ([Section 9.1](https://datatracker.ietf.org/doc/html/rfc3711#section-9.1)), although the above bounds
   are on a per stream (i.e., per SSRC) basis, the sender MUST base re-
   key decision on the stream whose sequence number space is the first
   to be exhausted.
```

I'll try to dig a bit in libwebrtc, see how they handle all of this

PR #184, opened 2022-03-24
Add support for AES128_CM_HMAC_SHA1_32

adamroach:
Adding support for SRTP_AES128_CM_HMAC_SHA1_32 as described in RFC 5764.
This necessitated some light refactoring of how `srtpCipher`s interact
with `ProtectionProfile`: the handling of this profile is identical
to handling for SRTP_AES128_CM_HMAC_SHA1_80, but the parameters are
different. This change assigns a profile to a cipher when it is created,
and all profile parameters are delegated to the profile (rather than the
previous setup, which split the values of profile parameters across the
profile and the cipher). This change also splits `authTagLen` into
`rtpAuthTagLen` and `rtcpAuthTagLen`, since they can have different
values depending on the profile in use.

### DIFF
--- a/context.go
+++ b/context.go
@@ -86,9 +86,9 @@ func CreateContext(masterKey, masterSalt []byte, profile ProtectionProfile, opts
 
 	switch profile {
 	case ProtectionProfileAeadAes128Gcm:
-		c.cipher, err = newSrtpCipherAeadAesGcm(masterKey, masterSalt)
-	case ProtectionProfileAes128CmHmacSha1_80:
-		c.cipher, err = newSrtpCipherAesCmHmacSha1(masterKey, masterSalt)
+		c.cipher, err = newSrtpCipherAeadAesGcm(profile, masterKey, masterSalt)
+	case ProtectionProfileAes128CmHmacSha1_32, ProtectionProfileAes128CmHmacSha1_80:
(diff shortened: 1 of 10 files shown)


Sean-Der:
@adamroach All the changes look great to me!

I added `TestProtectionProfileAes128CmHmacSha1_32` and it isn't passing. I am debugging now, see anything obvious?

```
--- FAIL: TestProtectionProfileAes128CmHmacSha1_32 (0.00s)
    srtp_test.go:527: failed to verify auth tag
FAIL
exit status 1
FAIL    github.com/pion/srtp/v2 0.006s
```

Sean-Der (approved):

Sean-Der:
@adamroach got it! `authKeyLen` was missing an entry.

Should we make this a default cipher suite in `pion/webrtc` after `dtls.SRTP_AEAD_AES_128_GCM, dtls.SRTP_AES128_CM_HMAC_SHA1_80` or do you just want this available but not on by default?

adamroach:
Sorry, I missed your messages yesterday. :)

Thanks for the merge and fixes! In terms of a default cipersuite for WebRTC, I wouldn't recommend adding this one in. 32-bit auth tags only make sense in some specific situations involving extremely low-bandwith audio codecs, sent over similarly low-bandwidth (or expensive) data links. So it's useful to have in certain non-WebRTC contexts, but I don't see it being useful for WebRTC.

PR #195, opened 2022-05-13
Minor fixes to SRTP tests

jech:
- Fix typo in TestValidPacketCounter
- Don't skip test for GCM
- Fix test for in-place encryption using GCM


davidzhao (approved):
lgtm!

PR #201, opened 2022-05-20
Optimise generateCounter (AES-CM-HMAC-SHA1)

adriancable:
Before
```
goos: darwin
goarch: amd64
pkg: github.com/pion/srtp/v2
cpu: Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
BenchmarkGenerateCounter
BenchmarkGenerateCounter-12      	97850202	        12.26 ns/op
```

After
```
BenchmarkGenerateCounter
BenchmarkGenerateCounter-12      	175902676	         6.383 ns/op
```

This updates a proposed code change from @kixelated back in 2019.

### DIFF
--- a/key_derivation.go
+++ b/key_derivation.go
@@ -48,16 +48,19 @@ func aesCmKeyDerivation(label byte, masterKey, masterSalt []byte, indexOverKdr i
 // -       passing through 65,535
 // i = 2^16 * ROC + SEQ
 // IV = (salt*2 ^ 16) | (ssrc*2 ^ 64) | (i*2 ^ 16)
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):

kixelated:
ez

kixelated:
I don't remember why this is faster, and it could probably be optimized further, but thanks for pushing this through.

adriancable:
On big endian systems you can do a couple of unsafe pointer casts to uint32 in the middle of the counter array instead. This would reduce op count by a further 2.5X. But at this point you are nibbling at the edges because the CPU footprint of this function is already very small now.

PR #206, opened 2022-05-23
Fix SRTCP packet length check

at-wat:
Packet length check of DecryptRTCP was wrong on AEAD profiles.


jeremija (approved):
Don't have a lot of context, but LGTM

PR #209, opened 2022-06-14
Fix: struct padding for 8 bytes alignment

Antonito:
This PR fixes the memory layout of some structs to reduce their size (aligned for 8 bytes, which should be most platforms).

- `srtpSSRCState` goes from `40` bytes to `32` bytes
- `ReadStreamSRTCP` from `56` to `48`
- `ReadStreamSRTP` from `56` to `48`

Those are small improvements, but should help in large scale scenarios.

I noticed this is an optimization that could be made across most pion repositories, if that's something worth doing @Sean-Der ?

---
If someone wants to go through this process, here's how I did it:

- Set the Go version as `1.18` in `go.mod`
- Add this file `memory_layout_helpers.go` to the package:
```go
package srtp

import (
	"fmt"
	"reflect"
	"runtime"
)

// BenchmarkAllocation allocates `amount` elements of type T, and display the memory usage.
//
// For ease of optimization, each member of the element T is printed, with its offset, padding & alignment.
func BenchmarkAllocation[T any](amount int) {
	typ := reflect.TypeOf(ZeroValue[T]())

	fmt.Printf("Struct is %d bytes long\n", typ.Size())

	n := typ.NumField()
	totalMinimumSize := uintptr(0)

	for i := 0; i < n; i++ {
		field := typ.Field(i)
		fmt.Printf("%-40s at offset %-8d | size = %-3d | align = %d\n",
			field.Name, field.Offset, field.Type.Size(),
			field.Type.Align())

		totalMinimumSize += field.Type.Size()
	}

	if typ.Size() != totalMinimumSize {
		fmt.Printf("with no padding, struct could be %d bytes long (%d bytes gain)\n", totalMinimumSize, typ.Size()-totalMinimumSize)
	}

	var allStats []T
	for i := 0; i < amount; i++ {
		allStats = append(allStats, ZeroValue[T]())
	}

	PrintMemoryUsage()
}

func PrintMemoryUsage() {
	var m runtime.MemStats
	runtime.ReadMemStats(&m)

	fmt.Printf("Alloc = %v MiB", bytesToMegabytes(m.Alloc))
	fmt.Printf("\tTotalAlloc = %v MiB", bytesToMegabytes(m.TotalAlloc))
	fmt.Printf("\tSys = %v MiB", bytesToMegabytes(m.Sys))
	fmt.Printf("\tNumGC = %v\n", m.NumGC)
}

func bytesToMegabytes(b uint64) uint64 {
	return b / 1024 / 1024
}

func ZeroValue[T any]() (value T) {
	return
}
```

- Add `memory_layout_test.go` to the package:
```go
// It is important to be in package `srtp` and not `srtp_test` to
// test internal structs.
package srtp

import (
	"testing"
)

func Test_StructMemoryLayout(t *testing.T) {
	BenchmarkAllocation[ReadStreamSRTP](1_000_000)
}
```
- Run the test!


adriancable:
@Antonito - can you quantify the improvement here in typical usage? (speed, memory usage, etc.)

My concern with this kind of thing is the possibility that some dependency may rely on the internal structure or ordering of structs. For example maybe there's a project out there using Pion that sends the structs directly over the wire without any serialisation. This PR would break that if different versions of Pion are used at each end of the wire, and it is quite hard to determine if there are any such usages out there. So, these changes are not risk-free, therefore, knowing the scale of the benefit is important in making a go/no-go decision.

Antonito:
Thanks for the detailed answer @adriancable !

I definitely understand your concerns, and to be quite frank had the same ones.
Here's what I concluded (feel free to push back if I'm mistaken!):
- `srtpSSRCState` is an internal state and holds a pointer (`replaydetector.ReplayDetector`), it can't be sent as-is over-the-wire.
- `ReadStreamSRTCP` and `ReadStreamSRTP` are public, however they both contain pointers as well (`*SessionSRTCP`, `chan bool`) and mutexes, so they can't been sent without any serialization over-the-wire as well.

Regarding improvements, this should simply reduce slightly memory usage on 64 bits platforms, and should not impact performances . 
It would mostly benefit SFUs with a large number of simultaneous connections (100k+).

If you want to push back that kind of changes to `v4.0.0`, it's also fine by me!

adriancable:
We're saving 8 bytes of memory per stream, maybe 24. If you have 100K+ connections, I would hope that you would not be running down to the last 800K - 2.4MB of RAM on the SFU!

Yes - every little helps - a little! And so I don't want this change to get forgotten, even if we don't merge it now. I think I need some second opinions on the right timing for this before pushing the button. @Sean-Der

Sean-Der:
@stv0g agree! Just missed this PR originally 

we should merge this and enable to struct align lint INO

stv0g (approved):

PR #211, opened 2022-06-17
Use speedy XOR from pion/transport

jech:
Now that we have a speedy version of xorBytes in pion/transport,
we might as well use it.  Benchmarks indicate an improvement
of 10% to 20% in CTR encryption speed for large blocks, depending
on the architecture.

Also remove TestXorBytesBuffer, which is no longer relevant.


go.mod
```diff
@@ -6,6 +6,6 @@ require (
 	github.com/pion/logging v0.2.2
 	github.com/pion/rtcp v1.2.9
 	github.com/pion/rtp v1.7.13
-	github.com/pion/transport v0.13.0
+	github.com/pion/transport v0.13.1-0.20220609023836-26aa3f6dca85
```
adriancable, line 9: I'm not sure if this is desirable. Do we need to wait for @Sean-Der to 'prod' pion/transport to v0.13.1 so we're referencing a release instead of a dated commit?

jech: Yes, I think so.

adriancable: I've now pushed the v0.13.1 tag to pion/transport. Can you try fixing up go.mod accordingly and we'll see if that works?

jech:
> I've now pushed the v0.13.1 tag to pion/transport. Can you try fixing up go.mod
> accordingly and we'll see if that works?

Done.

adriancable (approved):

PR #238, opened 2023-05-16
Adds support for AEAD_AES_256_GCM

digitalix:
#### Description

This adds support for AEAD_AES_256_GCM in srtp module. Later webrtc itself will have to be updated as well to enable support for AEAD_AES_256_GCM fully.


Sean-Der:
@at-wat @stv0g What do you think about removing the `go.mod` bumping. Maybe I am missing the value add?

Users on old versions of Go will think they can’t use these libraries. We aren’t using any new Go features though so no reason to block them?

Sorry If I am missing the benefit. Just want Pion to work for most people possible

stv0g:
My motivation for this check has been mainly harmonization between the various Pion modules.
In my opinion, we should also set clear expectations about which Go versions are supported.
Currently, we are only testing Go 1.19 and 1.20 which matches Go's support policy.
Hamonizing the go.mod version also clearly states which language features and standard packages we are allowed to use.
Currently, we have largely varying versions in the different modules which makes it hard to grasp which features we can use.

PR #259, opened 2024-02-20
Export unexported functions

potaga:
#### Description
For performance reasons, some projects need to bypass part of the standard pion path and access the SRTP session directly. This PR exports the required functions.


Sean-Der:
We are able to circumvent the SRTP APIs using [SettingEngine.BufferFactory](https://pkg.go.dev/github.com/pion/webrtc/v4#SettingEngine)

PR #277, opened 2024-06-29
Add support for Master Key Indicator

sirzooro:
This adds support for Master Key Indicator (MKI). It is used to select one of pre-configured SRTP/SRTCP encryption keys.

To use it, Context has to be created with MasterKeyIndicator option, it specifies MKI for master key and salt passed to CreateContext. Additional master keys/salts with their MKIs can be added using AddCipherForMKI. To remove MKIs, use RemoveMKI.

All MKIs must have the same length, and use the same length of master key and salt - they use the same crypto profile.

SRTP/SRTCP packets by default are encrypted using first key/salt/MKI. To select other key/salt/MKI, use SetSendMKI.

key/salt/MKI used for decryption are chosen automatically, using MKI sent in encrypted SRTP/SRTCP packet.

Reference encrypted packets used in tests are the same as generated by libsrtp2.


srtp.go
```diff
@@ -30,9 +30,19 @@ func (c *Context) decryptRTP(dst, ciphertext []byte, header *rtp.Header, headerL
 		}
 	}
 
-	dst = growBufferSize(dst, len(ciphertext)-authTagLen)
+	cipher := c.cipher
+	if len(c.mkis) > 0 {
```
at-wat, line 34: As the default master key is appended to `c.mkis`, this is always true even if `MasterKeyIndicator` option is not specified. `*1`
https://github.com/pion/srtp/blob/4bbc17f9768e73f06e8178e2fe16908d138b9b40/context.go#L93-L97

context.go
```diff
@@ -66,52 +71,99 @@ type Context struct {
...
+	}
+
+	err = c.AddCipherForMKI(c.sendMKI, masterKey, masterSalt)
 	if err != nil {
 		return nil, err
 	}
+	c.cipher = c.mkis[string(c.sendMKI)]
 
-	saltLen, err := profile.SaltLen()
+	return c, nil
```
at-wat, line 99: Due to `*1`, `c.mkis` should be cleared if `c.sendMKI` is `nil`

context.go
```diff
@@ -66,52 +71,99 @@ type Context struct {
...
+}
+
+// AddCipherForMKI adds new MKI with associated masker key and salt. Context must be created with MasterKeyIndicator option to enable MKI support.
+func (c *Context) AddCipherForMKI(mki, masterKey, masterSalt []byte) error {
+	if len(mki) != len(c.sendMKI) {
```
at-wat, line 104: I think it's better to return an error like `MKI is not enabled` if `len(c.sendMKI)` is zero

context.go
```diff
@@ -66,52 +71,99 @@ type Context struct {
...
+		if errOpt := o(c); errOpt != nil {
+			return nil, errOpt
+		}
+	}
+
+	err = c.AddCipherForMKI(c.sendMKI, masterKey, masterSalt)
 	if err != nil {
 		return nil, err
 	}
+	c.cipher = c.mkis[string(c.sendMKI)]
```
at-wat, line 97: Is it always safe to use string casted bytes as key? (e.g. against non-printable chars and multi-bytes chars)

option.go
```diff
@@ -71,3 +71,14 @@ type nopReplayDetector struct{}
 func (s *nopReplayDetector) Check(uint64) (func() bool, bool) {
 	return func() bool { return true }, true
 }
+
+// MasterKeyIndicator sets MKI for RTP and RTCP.
```
at-wat, line 75: It would be better to mention that `mki` passed to the option is a MKI for the initial master key and any MKIs added later must have same length.

srtcp.go
```diff
@@ -23,9 +23,10 @@ func (c *Context) decryptRTCP(dst, encrypted []byte) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	tailOffset := len(encrypted) - (authTagLen + srtcpIndexSize)
+	mkiLen := len(c.sendMKI)
+	tailOffset := len(encrypted) - (authTagLen + mkiLen + srtcpIndexSize)
 
-	if tailOffset < aeadAuthTagLen {
+	if tailOffset < aeadAuthTagLen+8 {
```
at-wat, line 29: What value is 8?

context.go
at-wat: Also, it might be better for readability to have MKI length as an integer variable

srtp.go
sirzooro: Fixed context creation, it should not add empty MKI to mki map.

context.go
sirzooro: Fixed context creation, it should not add empty MKI to mki map.

context.go
sirzooro: Added new error as suggested. Regarding passing MKI len as int var - in go array also has length associated, so no need to do this. I have added extra comment to make your point clearer.

context.go
sirzooro: Yes, it works fine.

option.go
sirzooro: Updated comment.

srtcp.go
sirzooro: This is some leftover from changes for my earlier PR to fix packet length validation. Removed.

at-wat (approved):
LGTM!

PR #278, opened 2024-07-06
Fix packet length validation

sirzooro:
Missing packet length validations could cause crash.


at-wat (commented):
The fix looks nice!
Could you add test cases which feed short packet and confirm it doesn't crash?

sirzooro:
Yes, I can add it.

sirzooro:
Added tests.

at-wat (approved):
LGTM

PR #286, opened 2024-07-17
Added support for NULL ciphers

sirzooro:
Added support for NULL ciphers. When they are used, created SRTP and SRTCP packets are authenticated only (no encryption).

Received SRTP/SRTCP packets are checked if their authentication tag is valid, and extra SRTP protocol fields are removed before returning then to application.

Fixed processing of SRTCP packets with E (encryption) bit cleared, previously duplicate check and tag validation was not performed, and whole packet was returned as-is (with extra fields) from decryptRTCP.

Use of NULL ciphers can be enabled independently for SRTP and SRTCP using SRTPNoEncryption and SRTCPNoEncryption options. They can be used with key exchange protocols which allows to configure them separately.

Added support for SRTP_NULL_HMAC_SHA1_80 and SRTP_NULL_HMAC_SHA1_32 cipher suites. They use key and salt of the same length as AES_CM_128 ones.

Added new tests to verify test vectors from RFCs.

Verified results with ones produced by libsrtp2. All matches except for AEAD SRTP ones - libsrtp2 encrypts packet instead of adding authentication tag only.


context.go
```diff
@@ -102,6 +107,26 @@ func CreateContext(masterKey, masterSalt []byte, profile ProtectionProfile, opts
 	return c, nil
 }
 
+// createContextWithCipher creates a new SRTP Context with a pre-created cipher. This is used for testing purposes only.
```
Sean-Der, line 110: If this is for tests only should it go in a `_test.go` file?

Sean-Der:
Amazing work @sirzooro thank you so much for doing this, this will make a lot of people happy :)

option.go
```diff
@@ -84,3 +84,37 @@ func MasterKeyIndicator(mki []byte) ContextOption {
 		return nil
 	}
 }
+
```
Sean-Der, line 87: Would it be possible to not have these options, and have everything configured just from Cipher choice?

I am just starting to read the PR, maybe I will understand better as I go :)

srtp_cipher.go
```diff
@@ -49,3 +49,13 @@ It is *before* the ESRTCP word (Encrypted-flag and SRTCP index).
 
 See https://tools.ietf.org/html/rfc7714 for the full specifications.
 */
+
```
Sean-Der, line 52: Maybe move to test file? Just to make it easier on readers in the future

context.go
sirzooro: Good idea, will move all test-related stuff to some _test.go file.

srtp_cipher.go
sirzooro: Good idea, will move all test-related stuff to some _test.go file.

option.go
sirzooro: Nope. MIKEY (RFC 3830) and `a=crypto` SDP attribute (RFC 4568) allows to disable encryption for SRTP and SRTCP independently. I also would have to create 3 additional crypto profiles for every cipher.

Maybe down the road it would be better to change ProtectionProfile to struct and put various additional params there, and pass it to `CreateContext` instead of options like it is implemented now.

sirzooro:
Moved new test-related code to srtp_cipher_utils_test.go file.

Sean-Der (approved):

PR #320, opened 2025-06-07
Fixes to use the same buffer as input and output

sirzooro:
Added checks for the same buffer to stop copying data when not needed. Also fixed srtpCipherAesCmHmacSha1 to work with one buffer when encrypting and decrypting RTCP packets.

### DIFF
--- a/srtcp.go
+++ b/srtcp.go
@@ -20,9 +20,13 @@ Simplified structure of SRTCP Packets:
 - Auth Tag - used by non-AEAD profiles only
 */
 
-const maxSRTCPIndex = 0x7FFFFFFF
+const (
+	maxSRTCPIndex = 0x7FFFFFFF
 
-const srtcpHeaderSize = 8
+	srtcpHeaderSize     = 8
+	srtcpIndexSize      = 4
+	srtcpEncryptionFlag = 0x80
+)
 
 func (c *Context) decryptRTCP(dst, encrypted []byte) ([]byte, error) {
 	authTagLen, err := c.cipher.AuthTagRTCPLen()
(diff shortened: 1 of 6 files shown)


srtcp.go
```diff
@@ -59,9 +63,7 @@ func (c *Context) decryptRTCP(dst, encrypted []byte) ([]byte, error) {
 		}
 	}
 
-	out := allocateIfMismatch(dst, encrypted)
```
at-wat, line 62: note: `allocateIfMismatch` handled nil `dst` but `growBuffer` in `cipher.decryptRTCP` also handles nil output buffer

srtp_cipher_aes_cm_hmac_sha1.go
```diff
@@ -195,81 +199,107 @@ func (s *srtpCipherAesCmHmacSha1) decryptRTP(
...
 	if subtle.ConstantTimeCompare(actualTag, expectedTag) != 1 {
 		return nil, ErrFailedToVerifyAuthTag
 	}
 
-	isEncrypted := encrypted[tailOffset]>>7 != 0
+	dst = growBufferSize(dst, decryptedLen)
+	sameBuffer := isSameBuffer(dst, encrypted)
+
+	if !sameBuffer {
+		copy(dst, encrypted[:srtcpHeaderSize]) // Copy the first 8 bytes (RTCP header)
```
at-wat, line 291: Looks like the original code didn't do this?

at-wat: `allocateIfMismatch` seems to have copied the buffer.
Fixed code is nice👍

at-wat (approved):
LGTM

PR #327, opened 2025-06-22
We never allocate more than 32 bytes AES/CTR

### LINKED ISSUE #322: Using a pool for IV in xorBytesCTR is probably counter-productive
In commit 099a0b0c209aa5907faea57ff7b1db2409555541, the allocation strategy for IVs has been changed to use a pool.  However, the pool uses fixed-size allocations of 1500 bytes, which is way oversized for the SRTP IV.

The code should be changed to use a smaller buffer size, aligned to the usual IV size, and to fallback to normal allocation for larger IVs.  Or it should simply be reverted.

@aankur, what problem are you trying to solve?  Could you please share your benchmarks results so I can decide whether to simply revert your change or implement a more complicated allocation strategy?

aankur:
#### Reference issue
Fixes #322

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -20,16 +20,18 @@ func incrementCTR(ctr []byte) {
 	}
 }
 
+const xorBufferSize = 32
+
 var xorBufferPool = sync.Pool{ // nolint:gochecknoglobals
 	New: func() any {
-		return make([]byte, 1500)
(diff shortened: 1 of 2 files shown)


Sean-Der:
I am so sorry this took so long @aankur lots of stuff has piled up lately, merging now!

Sean-Der (approved):

PR #352, opened 2025-11-13
ECN support

penhauer:
The `ReadStreamSRTP`, `ReadStreamSRTCP`, `SessionSRTP`, and `SessionSRTCP` have been changed with additional `ReadWithAttributes` methods and other needed methods. The important point is, `nextConn` 's type is now `transport.NetConnSocket`, which is a wrapper around `net.Conn`, where the new ReadWithAttributes method just behaves like Read. In the case where ECN is needed, `NewSessionSRTPWithNewSocket` should be used.

### DIFF
--- a/session.go
+++ b/session.go
@@ -146,7 +146,7 @@ func (s *session) start(
 		}()
 
 		b := make([]byte, 8192)
-		attr := transport.NewPacketAttributes()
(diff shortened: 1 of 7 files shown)


JoTurk:
replaced with the attributes API https://github.com/pion/srtp/pull/391

PR #391, opened 2026-09-16
Preserve packet attributes

JoTurk:
#### Description

Adds ReadWithAttributes to SRTP / SRTCP readers to read and preserve transport attributes.

https://github.com/pion/transport/pull/399


session.go
```diff
@@ -146,9 +147,10 @@ func (s *session) start(
 		}()
 
 		b := make([]byte, 8192)
+		var attrs packetio.Attributes
 		for {
 			var i int
-			i, err = s.nextConn.Read(b)
+			i, attrs, err = readWithAttributes(s.nextConn, b, attrs)
```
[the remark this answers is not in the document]
JoTurk: Doesn't really need to be tested here, It will be tested end to end in webrtc.

stream.go
```diff
@@ -3,6 +3,36 @@
...
+		return reader.ReadWithAttributes(buf, attrs)
+	}
+	clear(attrs)
+	n, err := reader.Read(buf)
+
+	return n, attrs[:0], err
+}
+
+func writeWithAttributes(writer io.Writer, buf []byte, attrs packetio.Attributes) (int, error) {
+	if writer, ok := writer.(interface {
```
mengelbart: Should this be an explicit interface like the attributeReader above?

mengelbart (dismissed):

stream.go
JoTurk: good catch, i moved the read one into an explicit interface after i tested the code, and i forgot to move the writer

JoTurk:
@mengelbart  your approval got dismissed because this repo has "Dismiss stale pull request approvals when new commits are pushed" i disabled it for consistently with other repos.

